PR #47, opened 2014-07-05
Using Spring Framework 4 Bill of Materials POM

arey:
See https://spring.io/blog/2013/12/03/spring-framework-4-0-rc2-available


michaelisvy:
Already merged.

PR #52, opened 2014-09-25
Dandelion without cdn

gkislin:
Switch all javascript and css to dandelion bandles (http://dandelion.github.io/dandelion/features/asset-bundles/) to eliminate datatables CDN in generated JSP.
Update jquery and jquery-ui.


michaelisvy:
Adding @tduchateau so we can get his opinion.

michaelisvy:
It seems that this solution will be much simpler to implement starting from Dandelion 1.0.0 (coming soon!). 
So let's wait for 1.0.0 to be out, no rush.

Issue #54, opened 2014-10-22
Petclinic doesn't work with Java 8


thomasdarimont:
The tomcat version 7.0.30 that is currently used doesn't work with Java 8.
If one tries to access `http://localhost:9966/petclinic`.

```
HTTP Status 500 - Unable to compile class for JSP:

org.apache.jasper.JasperException: Unable to compile class for JSP: 

An error occurred at line: 1 in the generated java file
The type java.util.Map$Entry cannot be resolved. It is indirectly referenced from required .class files

Stacktrace:
	org.apache.jasper.compiler.DefaultErrorHandler.javacError(DefaultErrorHandler.java:102)
```

Updating the maven tomcat-plugin and manually configuring the tomcat versions fixes the problem.

lukaseder:
I can confirm that @Scaronthesky's fix works for me as well, on Java 8u40

michaelisvy:
Hi,
my apology for the delay. For some reason my email notifications had been turned off. This should be fixed now, please let me know in case you still have any issue.

Cheers,
Michael.

PR #55, opened 2014-10-22
#54 - Update Tomcat versions to support running petclinic with Java 8.

thomasdarimont:
Updated used tomcat7 to latest versions to fix incompatibility with Java8.


michaelisvy:
Thanks Thomas, this is fixed now.

PR #57, opened 2014-12-11
asaassa

anshul-kapoor:


michaelisvy:
Thanks for your feedback.
I agree that Julien's blog is great. But it doesn't mean that we should include everything in Petclinic (which is meant for people to get started on Spring). 

This "stateless sessions" part is being discussed here: #77

Please feel free to raise some separate issues in case you think that some other parts are relevant for beginners. In any case, it's better to raise issues, not Pull Requests. Julien's PR is now 2 years old and our code has changed a lot in the meantime. We will need to apply changes manually.

PR #59, opened 2015-02-01
little changes

huguangju:
get rid of warnings


michaelisvy:
Thanks @huguangju for your great ideas. I couldn't accept your pull request because too many things have changed in Petclinic but I've implemented your ideas for LogBack and our CSS content.

PR #63, opened 2015-03-05
[#54] Petclinic doesn't work with Java 8

lukaseder:
See description in this issue here:
https://github.com/spring-projects/spring-petclinic/issues/54


michaelisvy:
This is fixed now.

PR #80, opened 2015-05-27
encodingFilter MUST be the first one to properly work.

liouxiao:
see http://forum.spring.io/forum/spring-projects/web/4782-how-to-set-setcharacterencoding-on-request

```
This filter should be FIRST filter in the filter-chain.
```

here're more references:

http://stackoverflow.com/questions/20863489/characterencodingfilter-dont-work-together-with-spring-security-3-2-0

http://forum.spring.io/forum/spring-projects/roo/90383-roo-1-1-0-release-problem-with-utf8-german-and-cyrillic


liouxiao:
Besides the encodingFilter order, I also add more i18n support as well as Chinese resource bundles.

liouxiao:
Michael,

I'm ok for the I18N part, but why not move the encoding filter to the first one (which should demonstrate the proper usage of this filter)?

aaguilera:
I agree with @liouxiao's suggestion about moving the encoding filter to be the first one.

michaelisvy:
Thanks for your feedback, I've moved the encoding filter to the top.

PR #93, opened 2015-09-17
Minor fixes - indentation mostly

trepel:
Fixes of indentation mostly, also few fixes of comments. Should not affect functionality in any way.


tduchateau:
Using EditorConfig (http://editorconfig.org/) could be a good idea to fix all indentation-related issues. I thing it would interest @michaelisvy and all other future contributors.

michaelisvy:
Hi, @arey and myself are the 2 main contributors for Petclinic. Since that project is pretty small, we haven't really agreed on indentation. Will discuss with him to see if that's something we can work with :).

Cheers,
Michael.

arey:
@tduchateau see my answer #96

PR #94, opened 2015-09-29
Add missing json-simple dependency

freemansoft:
This was originally a pull request with an entry in pom.xml.  The comments below made me revert that for the tree.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,10 +88,6 @@
 			<artifactId>taglibs-standard-jstlel</artifactId>
 		</dependency>
 		<!--  JSon -->
-	 	<dependency>
-		   <groupId>com.googlecode.json-simple</groupId>
-		   <artifactId>json-simple</artifactId>
-         </dependency>
 		 <dependency>
             <groupId>com.jayway.jsonpath</groupId>
             <artifactId>json-path</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -91,7 +91,6 @@
 	 	<dependency>
 		   <groupId>com.googlecode.json-simple</groupId>
 		   <artifactId>json-simple</artifactId>
-		   <version>1.1</version>
          </dependency>
 		 <dependency>
             <groupId>com.jayway.jsonpath</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,11 @@
 			<artifactId>taglibs-standard-jstlel</artifactId>
 		</dependency>
 		<!--  JSon -->
+	 	<dependency>
+		   <groupId>com.googlecode.json-simple</groupId>
+		   <artifactId>json-simple</artifactId>
+		   <version>1.1</version>
+         </dependency>
 		 <dependency>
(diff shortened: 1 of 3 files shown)


michaelisvy:
Thanks for that. I've actually hit the same issue this morning but I think it was a refresh issue in m2eclipse (because it works well when I launch Petclinic in the command line). 
datatables-core 1.0.1 actually comes with a transitive dependency to json-simple. We shouldn't need to declare it explicitly, unless I'm missing something?

freemansoft:
Its weird because 
"without the dependency" json-simple-1.1.1.jar shows up in the eclipse project dependency list but is not deployed as part of the application when testing in eclipse
"with the dependency without the version number" json-simple-1.1.1.jar shows up in the eclipse project dependency list and is deployed as part of the application

michaelisvy:
It really seems like an Eclipse refresh issue. 
If you deploy in the command line using 'mvn tomcat7:run' everything should be fine.

freemansoft:
Changed goal of this changelist to add a couple comments around aop invocation dependencies

freemansoft:
Closing pull request based on this discussion.  I made more changes on this branch so this will look weird when folks try and compare the code to the comments. 

Still learning to make each change on different branch or something with GIT

arey:
I've got the same issue but on IntelliJ when I've clicked on the "Find Owners" menu. Thus is not an Eclipse refresh problem.
Stacktrace:
```
org.apache.jasper.JasperException: javax.servlet.ServletException: java.lang.NoClassDefFoundError: org/json/simple/JSONValue
	org.apache.jasper.servlet.JspServletWrapper.handleJspException(JspServletWrapper.java:584)
	org.apache.jasper.servlet.JspServletWrapper.service(JspServletWrapper.java:454)
	org.apache.jasper.servlet.JspServlet.serviceJspFile(JspServlet.java:396)
	org.apache.jasper.servlet.JspServlet.service(JspServlet.java:340)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:725)
	org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
	org.springframework.web.servlet.view.InternalResourceView.renderMergedOutputModel(InternalResourceView.java:168)
	org.springframework.web.servlet.view.AbstractView.render(AbstractView.java:303)
```

The json-simple-dependency comes from dandelion-core.

michaelisvy:
thanks Antoine. Adding Thibault from Dandelion. @tduchateau 
Thibault, have you seen that before?

tduchateau:
@michaelisvy nop, but I'll have a look this evening.

michaelisvy:
I've checked on my side and I have exactly the same result as you guys: I need to include it as a transitive dependency otherwise it doesn't work (both on Eclipse and on the command line).

Also: I've noticed that both Spring Boot and Dandelion declare this dependency. I don't see why we would have a conflict still.

I've just committed an updated version of the POM with a temporary fix so people can run Petclinic.

PR #99, opened 2015-10-02
Removed HTTP PUT method - it is not supported in JSP 2.3

trepel:
In JSP 2.3 the HTTP PUT method is not supported. JSP 2.3 is used in Tomcat 8. Thus if deploying on Tomcat 8 the Spring Petclinic doesn't behave correctly if entering invalid values while updating pets or owners.

See http://stackoverflow.com/questions/23886941/http-status-405-jsps-only-permit-get-post-or-head

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/web/OwnerController.java
+++ b/src/main/java/org/springframework/samples/petclinic/web/OwnerController.java
@@ -116,7 +116,7 @@ public class OwnerController {
         return "owners/createOrUpdateOwnerForm";
     }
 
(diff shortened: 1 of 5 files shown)


arey:
I've got the same issue with Tomcat 8.0.18 when I edit the owner name with an empty value:
````
Etat HTTP 405 - JSPs only permit GET POST or HEAD
```
Instead of changing the controller @RequestMapping we may only change the JSP by using the _method=PUT parameter. The HiddenHttpMethodFilter is already set up in the web?xml
My question is why do we only have this error with invalid input?

PR #101, opened 2015-10-04
Fix issues when using the JDBC profile

mengelhardt:


arey:
Thanks @mengelhardt for your help.

PR #109, opened 2015-10-27
Updating Bootstrap version

dimitrisli:
From version 2.3.0 to 3.3.5


michaelisvy:
Thanks for that but I believe migrating to Bootstrap 3 has a bigger impact. We need to change many of the tags used in our JSP files.

arey:
Thanks @dimitrisli . We currently have a dedicated branch for the bootstrap3 update.

Issue #111, opened 2015-10-27
Setting dates doesn't work properly - they are shifted by one day


trepel:
Hi, when editing/storing entities with dates, the dates are shifted by one day back. I am not sure if you will be able to reproduce this. I came across similar issue in project using JSF and it was caused by TimeZone settings. My timezone is CET.

arey:
I've got the same behavior with GMT+1 timezone

michaelisvy:
hi @trepel , we are discussing about the new features we would like to have inside Petclinic. Since you have done quite a few pull requests, would you like to be involved? 
We were not sure how to contact you. If you'd like you can email me directly on misvy at pivotal dot io.

Michael.

PR #113, opened 2015-11-02
Using jodatime LocalDate instead of DateTime for visits

trepel:
This is very similar to https://github.com/spring-projects/spring-petclinic/commit/92de6557e10b610f3ea6579a6be67c53c4f766bc

The bug https://github.com/spring-projects/spring-petclinic/issues/111 also holds for visit dates - not only for pet birthdays - this should fix it.


arey:
Thanks @trepel

PR #130, opened 2016-02-07
Update .gitignore to exclude .metadata folder

ameya-pandilwar:


arey:
I'm sorry but it's not a best practice than a project and the eclipse workspace share the same directory: http://stackoverflow.com/questions/17402422/eclipse-workspace-location-best-practice

PR #132, opened 2016-02-19
added missing .mvn folder needed for maven wrapper

patrickcrocker:
I added the missing .mvn folder which is necessary for the mvnw and mvnw.cmd maven wrapper commands.


arey:
Thanks Patrick

PR #134, opened 2016-02-24
owners search has been case insensitive

broccolinisoup:
It is not an SQL standard. It works well with HSQLDB. However, mysql VARCHAR already allows case insensitive search. I mean, developers who use mysql for petclinic db, they already do the case insensitive searches. It is for just HSQLDB. So, we do not need to make any change on mysql/initDB.sql. I tested case insensitive search with both HSQLDB and MYSQL. Everything looks fine.

### DIFF
--- a/src/main/resources/db/hsqldb/initDB.sql
+++ b/src/main/resources/db/hsqldb/initDB.sql
@@ -36,7 +36,7 @@ CREATE INDEX types_name ON types (name);
 CREATE TABLE owners (
   id         INTEGER IDENTITY PRIMARY KEY,
   first_name VARCHAR(30),
-  last_name  VARCHAR(30),
+  last_name  VARCHAR_IGNORECASE(30),
   address    VARCHAR(255),
   city       VARCHAR(80),
   telephone  VARCHAR(20)


arey:
Do yo know if the VARCHAR_IGNORECASE type belongs to the SQL standard?
And have you test it on MySQL (mysql/initDB.sql)?

broccolinisoup:
Hello Arey,
Probably you are so busy I understand but the ticket is still waiting. Could you please check it? @arey

arey:
This ticket is still waiting because if your change is working on any RDBMS?

broccolinisoup:
Hello @arey ,
I wrote a comment but probably you haven't seen it. It is on the top.
It was "It is not an SQL standard. It works well with HSQLDB. However, mysql VARCHAR already allows case insensitive search. I mean, developers who use mysql for petclinic db, they already do the case insensitive searches. It is for just HSQLDB. So, we do not need to make any change on mysql/initDB.sql. I tested case insensitive search with both HSQLDB and MYSQL. Everything looks fine."

arey:
I'm sorry @broccolinisoup but I didn't see your first comment. In this condition, we can accept your change. Thanks you

broccolinisoup:
No worries @arey Thank you for your time!

Issue #137, opened 2016-03-05
Background from datapicker is 'transparent'


lukasz-szewc:
Screenshot of issue here:

http://postimg.org/image/hzpfbl2pf/

liouxiao:
The same issue in my place (happened in both Safari & Firefox running on Mac OS).

I add a CCS style in `petclinic.css` file to fix it:

```

.ui-datepicker {
	background-color: #d0e0c0;
}
```

arey:
I believe this issue is fixed ton the bootstrap3 branch. Could you please confirmed?

liouxiao:
Just confirm it, bootstrap3 branch has no this issue now. Thanks.

arey:
Thanks @liouxiao. So I close this issue.

PR #142, opened 2016-04-01
fix issue #137: Background from datapicker is 'transparent'

liouxiao:
as the title described.


arey:
Fix while Bootstrap 3 update

Issue #138, opened 2016-03-05
Css issue with 'width' 


lukasz-szewc:
Either header must be short and tables must fit in, or header should be extended.

Picture showing problem here:

http://postimg.org/image/azn640x5f/

arey:
@lukasz-szewc , I don't understand what is the issue? Do you propose to remove the width?

lukasz-szewc:
Please look at the header with menu. It is short and (presumably) should determine the borders of the page. Every content below should 'fit' within. This table violates that boundary, and whole page does not look symmetric at all. 

Here is another screen from app which in my opinion is correct:
I think that this is how this should look like:

http://postimg.org/image/mjw6e61h9/c7a66ef9/

Container class in css is quite clear about it: width: 700px;

lasiqueira:
Hello, I ran into the same issue. The CSS breaks when you enter the vetList.jsp or the ownerList.jsp.

I checked the code and chrome console and pinpointed what is the problem. 

I could fix that if you guys let me. I never contributed to an open source project before so I don't quite understand yet how to do it though.

arey:
Don't hesitate to submit a fix through a pull request.
For your information, I'm woking on Bootstrap 3 migration. Thus css will be updated. Maybe you may waithis update (planed for next week) then check if this issue is still there?

lasiqueira:
Oh, ok, I will wait then.

Thanks!

liouxiao:
actually this issue comes from `dandelion-datatables`, which inserts another `bootstrap.css` after `petclinic.css` in html head section, and the style file specifies `.container` width as `1170px` (in line 5833).

to quick fix it, we'd just specify the width at the html element in related jsp file, e.g. `ownerList.jsp`:

```
<body>
<div class="container" style="width: 700px">
...
```

lasiqueira:
Actually the best fix in my opinion is to remove the theme, since we already have bootstap in the application.

I created a pull request for that.

liouxiao:
@lasiqueira your fix looks more reasonable, thanks!

arey:
Could you tell me please if there is a similar issue on the bootstrap3 branch?
https://github.com/spring-projects/spring-petclinic/tree/bootstrap3

liouxiao:
In bootstrap3 branch, it looks by design the banner has a fixed width (600px), while the rest of page content has a full width, obviously there's no such issue anymore. (but maybe somebody would like the original layout design - "all content in a fixed width", :-)

PR #143, opened 2016-04-02
fix issue #138: Css issue with 'width'

lasiqueira:
Removed bootstrap2 theme from datatable since it was adding another
bootstrap.css

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/owners/ownersList.jsp
+++ b/src/main/webapp/WEB-INF/jsp/owners/ownersList.jsp
@@ -16,7 +16,7 @@
     <jsp:include page="../fragments/bodyHeader.jsp"/>
     <h2>Owners</h2>
 
-    <datatables:table id="owners" data="${selections}" row="owner" theme="bootstrap2"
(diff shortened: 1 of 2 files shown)


arey:
@lasiqueira could you check if this issue is still there on the bootstrap3 branch? As you, I've removed the bootstrap3 theme from database?

lasiqueira:
@arey There isn't anything in the /vendors/ folder, so it doesn't load any css and jquery...

arey:
In order to download all vendors resources, you have to execute a ```mvn install```

lasiqueira:
@arey I get this when I run mvn install:

Embedded error: Cannot run program "bower" (in directory "/home/luke/eclipse/jee-mars/eclipse/WS-petclinic/spring-petclinic"): error=2, No such file or directory

arey:
Are you running ```mvn install``` from Eclipse or from a terminal?
If we cannot resolve this problem, I may commit the vendors files.

lasiqueira:
@arey from a terminal. 

I was able to test it now that you have commited the vendors.

The CSS seems alright now!

PR #155, opened 2016-05-20
Fix logback + JMX memory leak on web application reload

markt-asf:
I've been using Spring PetClinic to test memory leaks in Apache Tomcat and I've found a memory leak in PetClinic related to logback and JMX. This pull request fixes the leak but, given I have next to no Spring development experience, I may well have gone about fixing this completely the wrong way.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,6 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <scope>runtime</scope>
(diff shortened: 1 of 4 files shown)


arey:
Thanks Mark for your contribution. To fix this memory leak, we decided to remove the JMX configurator. We would keep Petclinic as simple as possible. And adding a ServletContextListener is a little bit tricky for new Spring developers.
Feel free to submit other leaks or performance issues. And long life to Apache Tomcat.

PR #166, opened 2016-06-27
simplify jsp layout management

verydapeng:
abstract the page layout into “layout.tag”


arey:
In order to have both JSP pages for springboot and master branches, I've cherry-picked your change to the master.

arey:
In order to have both JSP pages for springboot and master branches, I've cherry-picked your change to the master.

PR #177, opened 2016-07-28
responsive menu fix

AndrejGajdos:
Responsive menu doesn't work. Data-target attribute added to fix that.


arey:
Thanks @AndrejGajdos

Issue #172, opened 2016-07-16
'mvn site' fails with cobertura-maven-plugin version 2.7


feststelltaste:
Trying to generate Maven site with 'mvn site' fails.

Exception:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.3:site (default-site) on project spring-petclinic: failed to get report for org.codehaus.mojo:cobertura-maven-plugin: Failed to execute goal org.codehaus.mojo:cobertura-maven-plugin:2.7:instrument (default) on project spring-petclinic: Unable to prepare instrumentation directory. source and destination are the same directory. -> [Help 1]
```

Possible workaround: 
- downgrading cobertura-maven-plugin to 2.6
- adding 'configuration/check/haltOnFailure' with value 'false' for the cobertura-maven-plugin

System:
- Oracle JDK 1.8.0_92
- Maven 3.3.9

PR #178, opened 2016-07-28
Downgrade Cobertura to enable 'mvn site'

feststelltaste:
Workaround as proposed in #172


arey:
Thanks for PR but I prefer to fix the issue in another way.

PR #182, opened 2016-08-29
Upgrade to Ehcache 3

snicoll:
I saw on twitter the reference of an article (in french):
http://javaetmoi.com/2016/08/migrer-vers-spring-boot/

That article concludes with something along the lines of  "Besides the
EhCache and Maven configuration, Petclinic does not hold a single line
of XML anymore".

Looking at the code, we can remove more XML if you want. This PR migrates
the cache infrastructure to EhCache 3 and JCache (JSR-107). This also
reduces the number of dependencies.


arey:
Thank you Stéphane. I have to update my article then report your PR to the master and/or the JavaConfig branch.

Issue #184, opened 2016-09-03
Make easier switch between MySQL and HSQLDB


arey:
To switch between MySQL and HSQDB databases, an improvment could be to add a Spring and a Maven profile.

PR #186, opened 2016-09-20
Make easier switch between MySQL and HSQLDB #184

bozzwin:
Signed-off-by: Mariusz Borwin <bozzwin@gmail.com>


arey:
Thank you @bozzwin

Issue #198, opened 2016-11-06
Distributed PetClinic with Spring Cloud


mszarlinski:
I would like to provide a "Spring Cloud" branch for the application to show how to split existing monolithic application into **microservices**. Distributed vesion would be based on a codebase from ["angularjs" branch](https://github.com/spring-projects/spring-petclinic/tree/angularjs) . What do you guys think?

arey:
I Maciej. This could be a great idea but could you please wait one or two weeks? Indeed the angularjs branch will be move to another repository.

mszarlinski:
Thanks for your replay. Actually I'm almost finished with a basic version built with Spring Cloud Config, Eureka, Hystrix and Zuul. Of course I can wait with a pull request for the new repository. By the time it is created, I should be ready with adjusted Docker setup. You can assign a task to me.

dsyer:
You might want to look at #200 as well. That kind of modularization could be useful. if you are going to split the "monolith" (otherwise I can't see the point of introducing all the Spring Cloud stack).

mszarlinski:
Hi Dave, thanks for your comment. I already did something similar with [Visit entity](https://github.com/mszarlinski/spring-petclinic/blob/spring-cloud/petclinic-visits-service/src/main/java/org/springframework/samples/petclinic/visits/domain/model/visit/Visit.java) when I was trying to define bounded contexts. And yes, I'm going to **split the monolith** :)

arey:
Dave, you may see the Maciej's work here: https://github.com/spring-petclinic/spring-petclinic-microservices
Now that the Spring Cloud Petclinic has its own repo, we could close this question.

mszarlinski:
@dsyer any comments and suggestions are certainly welcome :) I think after some time I will refresh microservices branch from this master (for example get rid of a thin service layer).

PR #200, opened 2016-11-11
Modularize and migrate to aggregate-oriented domain

dsyer:
Vet, Owner, Visit. The Visit "aggregate" is a little artificial
but it demonstrates a useful point about not holding on to
references of "parent" (reference data) objects, i.e. the Visit has
an Integer petId, instead of a Pet field. In principle this app is
now almost ready to migrate to multiple services if anyone wanted
to do that.


vfedoriv:
Hi.
As I see, you drop all service layer ?
IMHO, it's a not nice idea pull this in **master** branch.
Maybe, create new branch for microservices architecture ?

mszarlinski:
Hi @vfedoriv I am currently working on a microservices version and I named the branch "spring-cloud". Please see discussion in #198

dsyer:
Clarification: this is not a "microservice" conversion. This is just a modernization, using Spring Boot and a clear modular structure in a way that is more consistent with best practice in 2016, as opposed to 2003 when this app was first written.

vfedoriv:
"best practice" in Java - legacy support  ;)

vfedoriv:
Of course, you right. 
But half of this project "a little artificial". :)  How about model : BaseEntity -> NamedEntity ->... 
We really need this ?

dsyer:
I agree those classes would have to be factored out or their features duplicated if you were to split it up into actual physical modules. My judgement is that this is a small enough task that I don't mind deferring it (probably it never need happen in this repo). Worth a comment in the README though.

olivergierke:
Should the `VisitController` move into the `visit` package, too?

dsyer:
It's part of the "owner" UI, and as such it depends on the `PetRepository`. So moving it to the visit package creates a tangle.

olivergierke:
I wonder if it makes sense to invert the relationship between `Pet` and `Visit`. I'd argue managing pets and owners is useful on its own (a CRM like application). The notion of visits to the doctor is then coming on top of that. Owners and pets *without* visits makes sense. A visit without a pet — not so much.

`VisitRepository` sort of already works that way as it exposes a `findByPetId(…)`. This makes me feel like keeping the visits mapped on the `Pet` side is kind of obsolete and sort of breaking the abstraction direction.

So I'd rather make the visit an aggregate on it's own so that the `Visit` instances are created via the repository by being bound to a `Pet` (or it's identifier, I think even using `Pet` is fine here). That would allow us to move the URIs for view creation to the top level.

We could also go ahead and merge this as is so that we can create another PR with my suggested changes to see how that effects the structure of the codebase overall.

WDYT?

dsyer:
I played with various combinations and discarded them, but I'm not sure if that was one of them. Incremental changes after a merge also makes sense, if there are no objections to the change in principle (seems not).

olivergierke:
Okay, summarizing, I don't think anything in the proposed changes that really violates the fundamental goals of the ticket. So feel free to merge and I'll give it a spin on top for separate discussion.

dsyer:
Merged 83ff9a5

PR #212, opened 2016-11-30
Replace anonymous class with lambda

oscr:
Replaces an anonymous class with lambda equivalent


arey:
Now that minimal Java version is 8, it makes sense.

Issue #207, opened 2016-11-18
Owner null null


arey:
While editing a pet with invalid data, firstname and lastname of the owner are replaced by null: 
<img width="997" alt="owner-null-null" src="https://cloud.githubusercontent.com/assets/838318/20423029/2c8f44f6-ad6e-11e6-9d98-25d5bde05012.png">
Maybe introduce by the JSP to Thymeleaf migration?

PR #214, opened 2016-12-03
Fix incorrect glyphicon name

oscr:
Fixes UI bug and thereby making the warning sign glyphicon visible again.

In `layout.html` the glyphicon is incorrectly refered to as `th-warning-sign`. [Reference](http://getbootstrap.com/components/)


oscr:
Note: an example of the bug can be seen in the screenshot for #207.

PR #215, opened 2016-12-05
fix packages

ddd-workshop:


arey:
Seems to be related to #200

dsyer:
The `VisitController` is in the `owner` package to avoid a tangle (it needs to refer to owner stuff and vice versa). Alternative re-factorings are possible (@olivergierke had a suggestion)

PR #218, opened 2016-12-05
Fix date formatting

oscr:
This commit fixes an inconsistency in the date formatting in the project.

differentformatting

Notice how both `yyyy/MM/dd` and `yyyy-MM-dd` is used. However you can't insert `yyyy-MM-dd` 

rejectformatting

Since most places uses  `yyyy-MM-dd` I changed the remaining places to unify the date formatting.


arey:
I've tested your fix and it's working. Thanks @oscr

Issue #220, opened 2016-12-18
Issue with default cache configuration 


htynkn:
In `VetRepository` class, there is a cache named `vets`. 

In result of 6da37c30d3ba6fe915b6e61324a188b81cb2fb42, the configuration for `vets` is removed.

When you access ·vets· list, it will throw a cache exception.

http://localhost:8080/vets.html
![17661029453733](https://cloud.githubusercontent.com/assets/659135/21291193/5a924500-c511-11e6-8412-45674056fdfd.png)

swapy:
+1 this is indeed a issue.

anthonydahanne:
hello, Terracotta / ehcache developer here
apparently @snicoll added ehcache support in a programmatic way not so long ago (nice !) but his configuration got removed with this commit : https://github.com/spring-projects/spring-petclinic/commit/6da37c30d3ba6fe915b6e61324a188b81cb2fb42

I guess this commit intent was to get rid of a specific jsr-107 implementation; that makes sense.
But I believe instructions to re enable any implementation would help users. (maybe  using the [RI](https://github.com/jsr107/RI with some commented code samples to enable ehcache3)  ?

dsyer:
I believe it was fixed here: 80269539e270625. If you want to add a link in the README to the Spring Boot reference guide showing how to use JSR-107 that makes sense. I don't like having a load of commented out code in the app though.

anthonydahanne:
yes , sorry you're right that was fixed with https://github.com/spring-projects/spring-petclinic/commit/80269539e27062592f4c28b0d5498f38fbc2d578
I could run the pet clinic successfully using cache with : 
```
$ ./mvnw spring-boot:run -Dspring.profiles.active=production
[...]
2017-02-02 11:09:05.643  INFO 87239 --- [  restartedMain] org.ehcache.core.EhcacheManager          : Cache 'vets' created in EhcacheManager.
[...]
```
Thanks !

PR #223, opened 2016-12-26
Fix vets cache issue

kimbaudi:
Fix [issue #220](https://github.com/spring-projects/spring-petclinic/issues/220) by restoring "vets" cache.

fix-vet-cache


dsyer:
Already fixed in 80269539e2 I think. Thanks, though.

PR #225, opened 2017-01-12
Fix serialization of Vet

koniahin:
Hello!
Default configuration for Cacheable("vets") used serialization. Vet class has Serializable annotation, but his parent haven't. So, after deserialization we lost all data from parent fields. Also, we lost information from Specialty class for same reason.


dsyer:
Fixed in 80269539e2. Thanks anyway.

PR #229, opened 2017-02-16
Put Ehcache back

henri-tremblay:
This code will create a correct default Ehcache cache.

I've hardcoded the cache size since it's the way it used to be.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,16 @@
       <scope>runtime</scope>
     </dependency>
 
+    <!-- EhCache -->
+    <dependency>
+      <groupId>javax.cache</groupId>
+      <artifactId>cache-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.ehcache</groupId>
+      <artifactId>ehcache</artifactId>
+    </dependency>
+
     <!-- webjars -->
     <dependency>
       <groupId>org.webjars</groupId>
--- a/src/main/java/org/springframework/samples/petclinic/system/CacheConfig.java
+++ b/src/main/java/org/springframework/samples/petclinic/system/CacheConfig.java
@@ -1,14 +1,30 @@
 package org.springframework.samples.petclinic.system;
 
+import javax.cache.configuration.Configuration;
+import javax.cache.configuration.MutableConfiguration;
+
+import org.springframework.boot.autoconfigure.cache.CacheManagerCustomizer;
 import org.springframework.cache.annotation.EnableCaching;
-import org.springframework.context.annotation.Configuration;
+import org.springframework.cache.jcache.JCacheCacheManager;
 import org.springframework.context.annotation.Profile;
 
 /**
- * Cache could be disable in unit test.
+ * Cache could be disabled in unit test.
  */
-@Configuration
+@org.springframework.context.annotation.Configuration
 @EnableCaching
 @Profile("production")
-class CacheConfig {
+class CacheConfig implements CacheManagerCustomizer<JCacheCacheManager> {
+
+    @Override
+    public void customize(JCacheCacheManager cacheManager) {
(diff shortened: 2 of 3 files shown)


anthonydahanne:
LGTM
@snicoll what do you think about it ?

snicoll:
I don't like it. Using this feature together with `spring.cache.cache-names` property  is the lazy option for those who want to get started without knowing what they're doing. It's a convenience really and the only reason we support is is because `spring.cache.cache-names` is a generic option (it's not jcache specific).

I'd rather use a `JCacheManagerCustomizer` and create the caches using `addCache` (removing the `spring.cache.cache-names` in the process). If we go that route, I wouldn't use any of the ehcache specific stuff there either. The other option is `ehcache.xml`. The end goal is that a user can play with this and replace ehcache with another JCache implementation.

henri-tremblay:
Interesting. :-) I tried to use as much defaulting as I could. But in fact in my own code I generally don't do that. Let me push you another version then.

One thing though. JCache has limitations. The cache creation needs to be implementation dependant to get a meaningful one. The only one supported by JCache is `new MutableConfiguration()` which create a cache with an infinite heap store. That is rarely what you want. I can use an `ehcache.xml` file instead but for me it is still implementation dependant code.

WDYT?

snicoll:
> The end goal is that a user can play with this and replace ehcache with another JCache implementation.

I understand your perspective of showing how a cache is meant to be configured. Our perspective if we accept this PR is that a user can checkout the code and see what happens when they swap the JCache implementation. That's why `ehcache.xml` can be a good solution, you can still use whatever advanced configuration technique. With your current proposal, the code wouldn't compile anymore. With this proposal, it will complain the cache doesn't exist and the user can research how to configure it using the JCache provide they have chosen.

I know that `MutableConfiguration` is very limited. But maybe that sample app doesn't need any advanced stuff at all.

For the reasons explained above, I think `ehcache.xml` has my preference.

henri-tremblay:
Ok. I just pushed a new version closer to what you want. I also thought you preferred and xml free solution. To stay implementation independent, I can return a `new MutableConfiguration()` from `createCacheConfiguration`. You are right, this is just an example so it can be enough and will show where users should change it.

snicoll:
I am not in the camp of those who think XML is horrendous and shouldn't be used anywhere. Thanks for the update but you're still using Ehcache specific API in your change. What am I missing?

henri-tremblay:
Nothing :-) It was a version before seeing your comment. Here is one without any Ehcache code.

snicoll:
Merged, thanks. FYI, there is a dedicated `JCacheManagerCustomizer` for the JCache abstraction.

henri-tremblay:
Neat trick to return the customizer instead of implementing it. Thanks.

For the records, with an ehcache.xml it will have looked like this:

application.properties:
```
# Cache
spring.cache.jcache.config=ehcache.xml
```

CacheConfig.java:
```java
@Configuration
@EnableCaching
@Profile("production")
class CacheConfig {
}
```

ehcache.xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<config
    xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
    xmlns:jsr107='http://www.ehcache.org/v3/jsr107'
    xmlns='http://www.ehcache.org/v3'
    xsi:schemaLocation="
        http://www.ehcache.org/v3 http://www.ehcache.org/schema/ehcache-core-3.1.xsd
        http://www.ehcache.org/v3/jsr107 http://www.ehcache.org/schema/ehcache-107-ext-3.1.xsd">

    <service>
        <jsr107:defaults enable-management="false" enable-statistics="true"/>
    </service>

    <cache alias="vets">
        <key-type>java.lang.Object</key-type>
        <value-type>java.lang.Object</value-type>
        <resources>
            <heap unit="entries">100</heap>
        </resources>
    </cache>

</config>
```

PR #231, opened 2017-02-27
Move VisitController to visit package

mszarlinski:
... as it has been misplaced in `owner` package.


arey:
@mszarlinski, comments of the #215 gives you the reason

PR #275, opened 2017-10-21
Update

shnaps:
Update from source for fork


shnaps:
Sorry, wrong move from me.

PR #326, opened 2018-05-27
Enable content negotation via path extension to fix /vets.json

aidenchiavatti:
Content negotiation via path extension is disabled by default in Spring Boot. This was causing the /vets.json page to show XML instead of JSON. 

More info about content negotiation found here: [https://docs.spring.io/spring-boot/docs/2.0.0.RC1/reference/html/boot-features-developing-web-applications.html#boot-features-spring-mvc-pathmatch](https://docs.spring.io/spring-boot/docs/2.0.0.RC1/reference/html/boot-features-developing-web-applications.html#boot-features-spring-mvc-pathmatch)


bclozel:
I don't think we should re-enable this property, as we've flipped this default as a best practice for Spring applications. If Petclinic is about best practices, this would be counter-productive.

Instead, we have several solutions: 

1. we can split the `showResourcesVetList` into two endpoints (`"/vets.json"` and `"/vets.xml"`, each with a different `produces` condition)
2. just drop the file extension in the path pattern and let the client provide an accept header. It looks like this is consumed by a browser, so probably not the best choice
3. just have one `"/vets"` endpoint that produces a fixed format
4. configure the content negotiation manager to support query parameter based content negotiation, and have just a `"/vets"` endpoint that clients can query with `"/vets?format=json"` and `"/vets?format=xml"`

arey:
For information, in the Kotlin fork, by simplicity, we have declared two endpoints: 
```
@GetMapping("/vets.html")
 fun showHtmlVetList(

 @GetMapping("vets.json", produces = ["application/json"])
 @ResponseBody
  fun showJsonVetList(): 
```
We will align to the best choice.

snicoll:
> We will align to the best choice.

Thanks @arey - is there a way to sync issues when you discover them first? I'd appreciate an issue report in the main repo.

arey:
@snicoll I will do it. 
I believe this issue has been introduced by a Spring Boot 2 upgrade.

snicoll:
@arey Thanks. This change was introduced in `2.0.0.RELEASE`.

src/main/resources/templates/vets/vetList.html
```diff
@@ -26,8 +26,7 @@ <h2>Veterinarians</h2>
 
   <table class="table-buttons">
     <tr>
-      <td><a th:href="@{/vets.xml}">View as XML</a></td>
-      <td><a th:href="@{/vets.json}">View as JSON</a></td>
+      <td><a th:href="@{/vets}">View as XML</a></td>
```
Dovchiproeng: how about keeping both View as XML and View as JSON by append the format query param in the href.
```
 <td><a th:href="@{/vets?format=xml}">View as XML</a></td>
 <td><a th:href="@{/vets?format=json}">View as JSON</a></td>
```

Dovchiproeng: I like the option 4 by @bclozel which we can remain view as json instead of removing it

bclozel: I don't really get the goal behind that feature, besides showing that you can force a content-type without proper content negotiation. That approach is a bit outdated now.

If the goal is to keep that feature like this, the closest thing is indeed to use option 4).

aidenchiavatti:
Most recent commit matches option 2 from above, as I agree that it looks to be the 'best practice'. Is there a consensus on a solution?

src/main/resources/templates/vets/vetList.html
snicoll: I don't think we should keep that feature so I'll remove that link as part of the polish.

PR #330, opened 2018-07-05
Adapt JMeter script to recent changes

arey:


src/test/jmeter/petclinic_test_plan.jmx
```diff
@@ -135,7 +135,7 @@
           <stringProp name="HTTPSampler.response_timeout"></stringProp>
           <stringProp name="HTTPSampler.protocol"></stringProp>
           <stringProp name="HTTPSampler.contentEncoding"></stringProp>
-          <stringProp name="HTTPSampler.path">${CONTEXT_WEB}/vendors/jquery/jquery.js</stringProp>
```
snicoll: Specifying the version here doesn't sound right to me.

src/test/jmeter/petclinic_test_plan.jmx
```diff
@@ -135,7 +135,7 @@
...
+          <stringProp name="HTTPSampler.path">${CONTEXT_WEB}/webjars/jquery/2.2.4/jquery.min.js</stringProp>
```
snicoll: Specifying the version here doesn't sound right to me.

arey: I'm agree, the hard-coded jQuery version could be improve.
This step makes stress test the Webjars library: http://blog.ippon.fr/2013/03/13/improving-the-performance-of-the-spring-petclinic-sample-application-part-3-of-5/
By the way, it appears that cache-Control header is not set to static resource.
We may enable caching of static resources by setting the ```spring.resources.cache.cachecontrol.max-age``` property. Maybe in a ```production``` profile?

snicoll: Those looks unrelated to that specific issue. I am happy to discuss or revisit any of this in a separate issue though. 

Can you please update the PR so that the change is focused only to the reported issue?

arey:
Replacing /webjars/jquery/2.2.4/jquery.min.js by /webjars/jquery/jquery.min.js is working. No need to use regex to extract the jQuery version.

src/test/jmeter/petclinic_test_plan.jmx
arey: Sure. I've opened the #331

snicoll:
Thanks, that's now merged in `master`.

PR #334, opened 2018-07-12
Fix wrong link for CacheConfiguration

cruftex:


snicoll:
@cruftex good catch, thanks!

PR #335, opened 2018-07-12
fix comment about the cache size and clarify configuration

cruftex:
The comment `Create a cache using infinite heap.` is wrong.
The JCache standard only requires a cache to store at least "some" values to pass the TCK. It is not required to be `heap` or `infinite`. cache2k limits to 2000 entries by default and passes the TCK.

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/system/CacheConfiguration.java
+++ b/src/main/java/org/springframework/samples/petclinic/system/CacheConfiguration.java
@@ -7,10 +7,38 @@ import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
+/**
+ * Cache configuration intended for caches providing the JCache API. This configuration
+ * creates the used cache for the application and enables statistics that
+ * become accessible via JMX.
+ *
+ * <p>This is a minimal configuration that makes use of the JCache API to enable
+ * caching with any JCache implementation out of the box. The utilized JCache implementation
+ * is specified by the dependency in the {@code pom.xml}.
+ *
+ * <p>If the used cache implementation is configured via its specific and non standard mechanism,
+ * this example configuration might be redundant and can be removed or replaced by
+ * a programmatic configuration via the implementation specific methods.
+ *
+ * <p>In case no programmatic cache configuration is needed for the selected caching
+ * implementation this configuration class can be removed and the {@link EnableCaching}
+ * annotation can be put on the application class.
+ *
+ * @author Jens Wilke
+ */
 @Configuration
 @EnableCaching
 class CacheConfiguration {
 
+    /**
+     * A customizer that creates the the cache used by the application.
+     * This implies that the cache is not existent before the application start and the
+     * cache must be created. This is the default behavior of a JCache implemenation, in case
+     * no additional configuration is present and typically will create a transient in memory cache.
+     * If a persistent cache implementation is used, creating the cache on startup
+     * may lead to a failure, when the cache was created before. If the cache life time
+     * is controlled separately, the call of {@code createCache} must be removed.
+     */
     @Bean
     public JCacheManagerCustomizer petclinicCacheConfigurationCustomizer() {
         return cm -> {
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/samples/petclinic/system/CacheConfiguration.java
```diff
@@ -19,8 +19,10 @@ public JCacheManagerCustomizer petclinicCacheConfigurationCustomizer() {
     }
 
     private javax.cache.configuration.Configuration<Object, Object> cacheConfiguration() {
-        // Create a cache using infinite heap. A real application will want to use a more fine-grained configuration,
-        // possibly using an implementation-dependent API
+        // Enable statistics via the JCache programmatic configuration API.
+        // The actual default size limit of the cache is implementation dependent and cannot be modified
```
snicoll: I am not sure that's a valid alternative now you've removed the previous content. The intent of the original description was to mention that this was configuring a cache in-memory. 

I agree the "infinite heap" part is irrelevant. At the time I meant to focus on the fact that using EhCache's specific API allows you to restrict the size in heap and various other options. 

Can you please rephrase the description to focus on the original intent (that this configuration does not configure any sort of persistence for the cache).

src/main/java/org/springframework/samples/petclinic/system/CacheConfiguration.java
```diff
@@ -19,8 +19,10 @@ public JCacheManagerCustomizer petclinicCacheConfigurationCustomizer() {
...
+        // via the standard JCache API. To set more configuration parameters check out the configuration
+        // capabilities of the used caching implementation.
```
snicoll: That and the (yet limited) standard API.

cruftex:
@snicoll, from the PR annotation:

> The intent of the original description was to mention that this was configuring a cache in-memory.

That was your intent and this is in fact what is happening when you use EHCache (or Caffeine, or cache2k) without further configuration. But that intent is not and cannot be reflected in the configuration.

The configuration here is generic to any JCache, which may be in heap, off heap or persistent. The comments and implications must therefor make sense for every JCache implementation.

> Can you please rephrase the description to focus on the original intent (that this configuration does not configure any sort of persistence for the cache).

That's correct, but it does not configure an in heap cache as well ;)

I thought about mentioning that this yields an in memory cache together with EHCache dependency when no additional configuration is present, but explaining EHCache semantics in a generic JCache configuration file feels very weird. 

IMHO, it must be crisp and fit for every JCache. Thoughts?

snicoll:
> IMHO, it must be crisp and fit for every JCache. Thoughts?

Yeah I agree. So perhaps we can focus the description on the statistics part then and mention that additional configuration options are available in this API and the JSR-107 implementation API ?

snicoll:
> It's quite detailed now, and the documentation is more than the code. But I think that's how an example should look like ;)

Sorry but I disagree. That's very useful information for sure but I consider it overwhelming considering what that class does. It's not really the purpose of petclinic to explain how JCache works or some (rather advanced) concepts like what happens if a persistent cache is used.

If you have a couple of links that provides more information I am happy to revisit the PR and use those links instead.

Thoughts?

cruftex:
> It's not really the purpose of petclinic to explain how JCache works or some (rather advanced) concepts like what happens if a persistent cache is used.

Fair point. The stuff is important to know in the context, but this is not the place to do it in that detail.

> If you have a couple of links that provides more information I am happy to revisit the PR and use those links instead.

The JCache Spec is a PDF which is not directly linkable (licensing....). 

Let's keep the PR dangling. Maybe a few sentences make sense to add in the Spring documentation itself, which is, BTW, a great introduction to all sorts of programming concepts. Then we link there.

> Thoughts?

Studying the Spring code, I think the best  would be to remove the need of `createCache` at all and make it default in the Spring JCacheManager. Like so:

````
	@Override
	protected Cache getMissingCache(String name) {
		CacheManager cacheManager = getCacheManager();
		javax.cache.Cache<Object, Object> jcache = 
                   cacheManager.createCache(name, new MutableCacheConfiguration());
		return new JCacheCache(jcache, isAllowNullValues());
	}
````
This would make the whole class redundant.

I think it would be great to be able to use the JCache caches as catch all as well.

Of course, now there are the backwards compatibility issues to consider....

snicoll:
We've had that discussion before and I can't find back the thread but creating the caches on the fly is not happening. We have to fail when a cache is not present so that users know they have to do something. And creating something with a default config doesn't sound a sane choice to me.

This thread is starting to expand to various topics and I'd like it to be focused. I am happy to review a PR to improve the Spring Framework documentation as long as it stays focused on what we do.

cruftex:
> This thread is starting to expand to various topics and I'd like it to be focused.

Thanks for the explanation and sorry for veering off. Maybe the typical "warmup phase".

Since this got too lengthy, I am closing this PR for now and will do a fresh one, once I have the time and gained more oversight.

snicoll:
No worries. Do you mind if I take bits and pieces of your PR to improve that class. It certainly would be better than the current situation!

cruftex:
Maybe that's the faster option.

You could also accept the PR and then strip it as you like. I didn't mean to do a hostile take over of this class ;)

snicoll:
Thanks a lot @cruftex ! I am happy to continue the discussion on the Spring Framework's tracker for further ideas to improve the doc.

PR #338, opened 2018-07-26
add dockerfile and instructions

paulczar:


pivotal-issuemaster:
@paulczar Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=338)!

snicoll:
@paulczar thanks for the PR. While we have instructions to run docker for services required by petclinic (e.g a database), I am not keen to add instruction to run petclinic with docker as they are much easier ways to do so.

PR #344, opened 2018-08-12
Added some asserts on PetclinicIntegrationTests

jabrena:
I was reviewing the Project example `Pet Clinic` and I discovered that one test doesn´t include any assert so I added some asserts.

Juan Antonio

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java
@@ -17,6 +17,7 @@ package org.springframework.samples.petclinic.model;
 
 import javax.persistence.Column;
 import javax.persistence.MappedSuperclass;
+import java.util.Objects;
 
 
 /**
@@ -45,4 +46,17 @@ public class NamedEntity extends BaseEntity {
         return this.getName();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof NamedEntity)) return false;
+        NamedEntity that = (NamedEntity) o;
+        return Objects.equals(name, that.name);
+    }
+
+    @Override
+    public int hashCode() {
+
+        return Objects.hash(name);
+    }
 }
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@jabrena Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=344)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=344) to manually synchronize the status of this Pull Request.

src/main/java/org/springframework/samples/petclinic/vet/Vet.java
```diff
@@ -15,11 +15,7 @@
  */
 package org.springframework.samples.petclinic.vet;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
dsyer: We never use "*" imports in Spring projects. Probably best to maintain that rule here.

src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java
```diff
@@ -45,4 +46,17 @@ public String toString() {
         return this.getName();
     }
 
+    @Override
```
dsyer: Are these new methods actually used anywhere?

src/main/java/org/springframework/samples/petclinic/vet/Vet.java
```diff
@@ -76,4 +73,17 @@ public void addSpecialty(Specialty specialty) {
         getSpecialtiesInternal().add(specialty);
     }
 
+    @Override
```
dsyer: I know that it can be a good idea to implement `equals()` and `hashCode()` for entities. Not sure these implementations are good enough to show best practice. Are they actually used in the app?

src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java
snicoll: @dsyer the updated test relies indirectly on `equals`. IMO, I am with you that the implementation isn't good enough to show best practices so I am rather keen to decline this. 

Having said that, the current test that this PR changes is problematic so we should do something about it.

PR #360, opened 2018-10-10
Migrate tests to JUnit 5

arey:


pom.xml
```diff
@@ -103,6 +110,18 @@
     </dependency>
     <!-- end of webjars -->
 
+    <!-- Testing -->
```
dsyer: You wouldn't need these with Spring Boot 2.2. Maybe we should make this a 2.2 change?

The exclusion above is unnecessary then as well.

arey: Fine, it will simply the XML configuration.
If you want, to remember it, I can create an issue in prevision of the Spring Boot 2.2 upgrade.

dsyer: Thanks. That would be great.

arey:
@dsyer I've traced this into the issue #422

arey:
Thank you Stéphane

pom.xml
travi1812: spring boot enquiry #2460 #2457

PR #361, opened 2018-10-10
Add reference to the Spring Data JDBC version of Spring Petclinic

arey:


readme.md
```diff
@@ -110,6 +110,7 @@ there.
 | [spring-petclinic-graphql][]       | GraphQL version based on React Appolo, TypeScript and GraphQL Spring boot starter |
 | [spring-petclinic-kotlin][]        | Kotlin version of [spring-petclinic][] |
 | [spring-petclinic-rest][]          | Backend REST API |
+| [spring-petclinic-data-jdbc][]     | The Spring Data JPA persistence layer of [spring-petclinic][] has been replaced by Spring Data JDBC |
```
snicoll: I'd like that we remove this table entirely and leave the link to the org. That way, the content is managed on the org as it should be.

It feels odd that we have to have an exhaustive list of something that is managed in a different org.

Leaving this open to see what @dsyer thinks.

dsyer: Agree. It makes much more sense to manage it over there. The only problem is that an org doesn't really have a README, so there isn't one obvious place to put this list that isn't here. Or maybe there's a Github feature I'm not aware of.

snicoll: We can update the link above when @arey has figured out where that content should live.

arey: I've done some research. GitHub offers to host organization website: https://help.github.com/articles/user-organization-and-project-pages/  
I've already created the Git repository: https://github.com/spring-petclinic/spring-petclinic.github.io
URL of the Spring Community website should be: https://spring-petclinic.github.io/
Now, I need some time to create the site.
Do you know if I could use the Spring/Pivotal bootstrap theme? Is there any copyright?

snicoll: > Do you know if I could use the Spring/Pivotal bootstrap theme? Is there any copyright?

I am not sure what you had in mind but, as the spring-petclinic is a separate org from the official/main repo, I prefer that we don't increase the confusion. I understand that this can be a lot of work though so perhaps we could find some sort of middle ground? You could have a `spring-petclinic` project with only the README for instance and we could point to that.

arey: @snicoll I've created a website for Spring PetClinic org. All forks are listed into the https://spring-petclinic.github.io/docs/forks.html page. Every commiter could contribute to the page: https://github.com/spring-petclinic/spring-petclinic.github.io/blob/master/docs/forks.md

snicoll: Thanks @arey I've added a link in https://github.com/spring-projects/spring-petclinic/commit/3b36b80741187865ddd4347b7cc13b1b3c041d51

snicoll:
As discussed in the comment, we've removed the table and add a link to the spring petclinic org instead.

arey:
Yes Stéphane, we could replace the table by a link to the spring petclinic org.


Le 22 octobre 2018 à 15:00:56, Stéphane Nicoll (notifications@github.com) a écrit:

As discussed in the comment, we've removed the table and add a link to the spring petclinic org instead.

Issue #161, opened 2016-06-18
Setup Jenkins / SonarQube


arey:
Analyzing source code of the Petlinic application into the Spring.IO SonarQube instance: https://sonar.spring.io/

arey:
Question has been aked by email to Stéphane Nicoll and Brian Dussault. I'm waiting their answer.

CMYanko:
I'm working on a jenkinsfile but struggling to get the build to work. I thought I could just run mvnw spring-boot:repackage but no joy. Also looking for how to change the port as 8080 is spoke for on my system.

snicoll:
@CMYanko have you had a look to [the readme](https://github.com/spring-projects/spring-petclinic#running-petclinic-locally) yet?

snicoll:
@ElizaSellegi petclinic is a spring boot application so please refer to its doc to [know how to change the port](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-customizing-embedded-containers). This is also hardly related to this very issue, you can use StackOverflow for questions.

snicoll:
@arey I am tempted to close this one as it didn't get much traction. Am I missing something?

arey:
Do you know any people at Pivotal who managed the https://sonar.spring.io/ platform? He will just have to declare the `spring-petclinic` repository.
It could make sense to prove the high quality code of this sample.

PR #383, opened 2019-01-13
Delete redundant sonar-project.properties file

mustafau:
Values in that file are wrong.


snicoll:
There is a discussion about SonarQube in #161 so removing the file makes sense. Thank you.

PR #385, opened 2019-01-30
Add support for Gradle

Am-Coder:
Attempting to migrate from maven to gradle build


pivotal-issuemaster:
@Am-Coder Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=385)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=385) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Am-Coder Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=385)!

snicoll:
> Attempting to migrate from maven to gradle build

I think the plan is that we're considering adding Gradle support so removing Maven is not an option.

Am-Coder:
I think its done.

.travis.yml
```diff
@@ -1,2 +1,4 @@
 language: java
 jdk: oraclejdk8
+before_install:
+  - chmod +x gradlew
```
dsyer: Is this really necessary? It would be better to commit it with the right flags if possible (presumably it works for mvnw).

build.gradle
```diff
@@ -0,0 +1,51 @@
...
+        url = 'https://repo.spring.io/milestone'
+    }
+
+    maven {
+        url = 'http://repo.maven.apache.org/maven2'
+    }
+}
+
+dependencies {
+    compile 'org.springframework.boot:spring-boot-starter-actuator:2.1.1.RELEASE'
```
dsyer: We should use the dependency management features in Spring Boot - no need for explicit versions in `build.gradle`.

foxundermoon:
`.gradle `  should ignore

.travis.yml
Lewiscowles1986: Is it really so important to not be explicit?

build.gradle
Lewiscowles1986: perhaps this is mentioned in https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-build-systems.html#using-boot-gradle

phstudy:
This PR does not include wro4j plugin to handle static resources generation. So this PR is not complete.

build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    maven {
+        url = 'https://repo.spring.io/snapshot'
+    }
+
+    maven {
+        url = 'https://repo.spring.io/milestone'
+    }
+
+    maven {
+        url = 'http://repo.maven.apache.org/maven2'
```
longlg88: I think it would be update to this URL.
```
url = 'https://repo.maven.apache.org/maven2'
```

Lewiscowles1986:
@dsyer is there any background on the re-opening?

dsyer:
It got closed by mistake (automatically by GitHub). There’s no actual change in status.

dsyer:
Obsoleted by f015573 (we don't need wro4j any more), but thanks for starting the ball rolling.

PR #450, opened 2019-09-22
Added jquery data table to list

DineshLakmal:


snicoll:
Thank you for the PR @DineshLakmal but we prefer to keep this version of petclinic as simple as possible on the client-side as the goal is to focus primarily on the backend. There is an org with other "variants" of petclinic, see https://github.com/spring-petclinic

PR #452, opened 2019-09-24
fix docs for setting profiles

faxx1080:
Add using -Dspring-boot.run.jvmArguments to the notes for how to set up using MySQL.

It seems newer versions of Spring may fork the JVM that runs the webserver, and this causes issues with passing properties like -Dspring.profiles.active. This advises a fix.

### DIFF
--- a/src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
+++ b/src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
@@ -22,15 +22,14 @@
    scripts (or set "spring.datasource.initialization-mode=always" the first time you run the app).
 
 3) Run the app with `spring.profiles.active=mysql` (e.g. as a System property via the command
-   line, but any way that sets that property in a Spring Boot app should work).
+   line, but any way that sets that property in a Spring Boot app should work). For newer versions
+   of Spring Boot, those will fork the JVM that executes the webserver, which means properties on
+   the command line are lost. Use
    
-4) For newer versions of Spring Boot, those will fork the JVM that executes the webserver,
-   which means properties on the command line are lost. You can use:
-    
    mvn spring-boot:run\
-   -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=mysql -Dspring.datasource.initialization-mode=always"
-    
-   To set profiles & properties.
+   -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=mysql"
+
+   To make mysql work with forking.
 
 N.B. the "petclinic" database has to exist for the app to work with the JDBC URL value
 as it is configured by default. This condition is taken care of by the docker-compose
--- a/src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
+++ b/src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
@@ -23,6 +23,14 @@
 
 3) Run the app with `spring.profiles.active=mysql` (e.g. as a System property via the command
    line, but any way that sets that property in a Spring Boot app should work).
+   
+4) For newer versions of Spring Boot, those will fork the JVM that executes the webserver,
+   which means properties on the command line are lost. You can use:
+    
+   mvn spring-boot:run\
+   -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=mysql -Dspring.datasource.initialization-mode=always"
+    
+   To set profiles & properties.
 
 N.B. the "petclinic" database has to exist for the app to work with the JDBC URL value
 as it is configured by default. This condition is taken care of by the docker-compose


pivotal-issuemaster:
@faxx1080 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=452)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=452) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@faxx1080 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=452)!

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
```diff
@@ -23,7 +23,15 @@
 
 3) Run the app with `spring.profiles.active=mysql` (e.g. as a System property via the command
    line, but any way that sets that property in a Spring Boot app should work).
+   
+4) For newer versions of Spring Boot, those will fork the JVM that executes the webserver,
```
snicoll: I don't think a new item should be added, considering that setting the version always worked. It just happend to work with only `-D` if you don't fork the process.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
```diff
@@ -23,7 +23,15 @@
...
+   which means properties on the command line are lost. You can use:
+    
+   mvn spring-boot:run\
+   -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=mysql -Dspring.datasource.initialization-mode=always"
```
snicoll: There is a dedicated property to set a profile, `- Dspring-boot.run.profiles=mysql`. I don't think we should mention datasource initialization again (as it should only be invoked the first time you run the app).

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
```diff
@@ -23,7 +23,15 @@
...
+    
+   To set profiles & properties.
 
 N.B. the "petclinic" database has to exist for the app to work with the JDBC URL value
 as it is configured by default. This condition is taken care of by the docker-compose
-configuration provided, or by the `schema.sql` if you can run that as root.
\ No newline at end of file
+configuration provided, or by the `schema.sql` if you can run that as root.
```
snicoll: Looks like an unrelated change, please revert.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
faxx1080: I reworded this into step 3.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
faxx1080: good point.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
faxx1080: Github's web editor added a newline at the end of the file automatically. Fixed.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
```diff
@@ -22,7 +22,14 @@
    scripts (or set "spring.datasource.initialization-mode=always" the first time you run the app).
 
 3) Run the app with `spring.profiles.active=mysql` (e.g. as a System property via the command
-   line, but any way that sets that property in a Spring Boot app should work).
+   line, but any way that sets that property in a Spring Boot app should work). For newer versions
+   of Spring Boot, those will fork the JVM that executes the webserver, which means properties on
+   the command line are lost. Use
+   
+   mvn spring-boot:run\
+   -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=mysql"
```
snicoll: That still doesn't use the dedicated attribute I've shared previously.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
```diff
@@ -22,7 +22,14 @@
    scripts (or set "spring.datasource.initialization-mode=always" the first time you run the app).
 
 3) Run the app with `spring.profiles.active=mysql` (e.g. as a System property via the command
-   line, but any way that sets that property in a Spring Boot app should work).
+   line, but any way that sets that property in a Spring Boot app should work). For newer versions
```
snicoll: As I've stated previously, I don't think we should add a note about "newer version". The attribute I've shared with "old" and "new" version irrespective of forking and is the recommended way to use the plugin. I would rather remove the content in parenthesis completely as it implies you can pass a system property that way and it will work.

src/main/resources/db/mysql/petclinic_db_setup_mysql.txt
```diff
@@ -22,7 +22,14 @@
...
+   of Spring Boot, those will fork the JVM that executes the webserver, which means properties on
+   the command line are lost. Use
+   
+   mvn spring-boot:run\
+   -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=mysql"
+
+   To make mysql work with forking.
```
snicoll: Not sure why this sentence was added or why you should mention forking. That property works with or without forking.

faxx1080: Would it be better to just have (3) say:
```
3) Run the app with `spring.profiles.active=mysql` (e.g. as a System property via the command
   line, but any way that sets that property in a Spring Boot app should work, like
   -Dspring-boot.run.profiles=mysql)
```

snicoll: I don't think so as it's maven specific and there's no connection to it the way you're suggesting

faxx1080: I'm not sure what would be better to write.

I thought of amending the docs because I recently wanted to run this using the provided
`./mvnw spring-boot:run`, then wanted to switch to mysql, and tried  `./mvnw -Dspring.profiles.active=mysql spring-boot:run` and `./mvnw spring-boot:run -Dspring.profiles.active=mysql`, and neither worked. I thought if for a beginner, the documentation showed a way to fix that issue, it could be helpful.

snicoll: Absolutely and your proposal is very much appreciated. Let me take over and I'll polish it if that's ok with you?

faxx1080: sure, thanks!
My apologies I'm not that familiar (yet) with your way of doing things.

snicoll: No need to apologize for anything. This back and forth is to be expected and not indicative of a problem. Thanks again for your contribution.

dsyer:
Cherry picked in b91e213

PR #514, opened 2020-01-03
Make gitpod initialize on startup

dsyer:


dsyer:
I think Codespaces makes this obsolete now that everyone can use it.

hogah97323:
@dsyer Is Codespaces free?

dsyer:
Hmm maybe not. If is for me but maybe individuals with no special org permissions can’t see it yet. I think Microsoft plan to open it up but maybe not yet? Let us know what your experience is.

hogah97323:
@dsyer Where have you seen the plan of open sourcing it? I currently can't have access to the Codespaces buttons.

PR #525, opened 2020-02-24
Create workflow.yml

N-Usha:
Hi,

Firstly great app for demonstrating Spring Boot framework. Kudos! :)
I have added a CI/CD workflow to build and deploy this app to Azure using GitHUb Actions. And wanted to contribute the same to the original repo. to make this work, you only would need to  [copy the Publish profile of the app](https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md#set-secret-with-web-app-publish_profile) and set it as a GitHub secret "AZURE_WEBAPP_PUBLISH_PROFILE" for the repo.  Request you to kindly accept this contribution as it may be useful for all those users to easily get started with GitHub Actions.

Thanks,
Usha

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,39 @@
+name: Deploy Java app to Azure Web App
+
+on:
+  [push]
+
+# CONFIGURATION
+# For help, go to https://github.com/Azure/Actions
+#
+# 1. Set up the following secrets in your repository:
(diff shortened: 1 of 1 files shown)


snicoll:
>  it may be useful for all those users to easily get started with GitHub Actions.

Thanks for the PR but the focus of this project isn't to help users getting started with GitHub Actions.  There would be an ongoing maintenance in merging this which I think isn't justified. Thanks anyway.

PR #567, opened 2020-03-16
pull request 2

khalilalturk:


snicoll:
This looks like a work assignment that shouldn't lead to the creation of a Pull Request to this repository. Could you please reach out to the person responsible so that they clarify this part of the assignment? This creates unnecessary noise to this repository. Thanks.

PR #668, opened 2020-09-01
Fix bug with replacing entities to DTO objects that were used as argu…

mahdieha:
Fix bug with replacing entities to DTO objects that were used as arguments of "@RequestMapping" methods on the controller.


akashsolanki:
@mahdieha There is a build failure with this commit. Please check.

snicoll:
@mahdieha I am not keen to use model mapper in this project and I am not clear what bug this PR is fixing. There isn't a test that demonstrates what was changed. Can you please clarify?

PR #673, opened 2020-09-04
Feat/672

mahdieha:
Added Lombok library and implemented it for entities.


src/main/java/org/springframework/samples/petclinic/model/BaseEntity.java
```diff
@@ -30,20 +33,15 @@
  * @author Juergen Hoeller
  */
 @MappedSuperclass
+@Data
```
ramilS, line 36: it's a bad idea to use @Data for Entity

snicoll:
@mahdieha thanks for the PR but we don't want to use Lombok.

Issue #682, opened 2020-10-01
First letter of "Pet Type" should be capital


anantjain6:
First letter of "Pet Type" should be capital

![](https://user-images.githubusercontent.com/22341750/94831915-eb6f2780-042a-11eb-8ecc-c1ba198cf712.png)

somayaj:
Hi can I contribute? @somayaj is my handle

anantjain6:
@somayaj i have already submitted pull request for it

snicoll:
Thanks for the suggestion @anantjain6.

Closing in favour of PR #683

PR #683, opened 2020-10-01
Capitalize the first letter of pet type

anantjain6:
Patch for #682


snicoll (changes requested):
Thanks but there are data.sql for other databases that should be upgraded too.

jaDEVirek (approved):

dsyer:
I guess no-one cared that much. Maybe better implemented as an i18n message anyway?

PR #685, opened 2020-10-03
Capitalized pet names 

viswa08:
Captitalisation of pet names has been done in database and are reflecting in UI.

pet-Type


snicoll:
Thanks but someone already raised a PR before you did, see #683.

Issue #667, opened 2020-09-01
Replace persistence entities with simple DTO objects.


mahdieha:
If we using a persistence object (@Entity) as an argument of a method annotated with @RequestMapping, it is possible that the end-user can change the content of unexpected fields in the database, which eventually lead to a critical bug. 

For example on the below controllers, we need to replace entities with a simple Dto:
1. OwnerController
2. PetController
3. VisitController

arey:
SonarQube references this security flaw: https://rules.sonarsource.com/java/tag/spring/RSPEC-4684

We have encountered this issue on other forks of Spring Petclinic (ie. https://sonarcloud.io/project/issues?id=spring-petclinic_spring-framework-petclinic&open=AWq4K0FjEGMyMe6L0asK&resolved=false&types=VULNERABILITY)

pedsf1968:
> If we using a persistence object (@entity) as an argument of a method annotated with @RequestMapping, it is possible that the end-user can change the content of unexpected fields in the database, which eventually lead to a critical bug.
> 
> For example on the below controllers, we need to replace entities with a simple Dto:
> 
> 1. OwnerController
> 2. PetController
> 3. VisitController

Yes @mahdieha for me we must isolate @Controller from @Repository by using @Service and DTO.
Controller call Service using DTO and Service call Repository using @Entity

pedsf1968:
I have send a pull request with the feature branch 
Feature/implement services

dsyer:
We long ago decided not to do this, purely in the interests of keeping the codebase as small as possible. While it is certainly an interesting idea, and can be seen as best practice in some situations, blanket use of DTOs for all entities in all settings is probably an anti-pattern. So I don't see any reason to change the decision to keep it simple. Maybe a link to some nice article on DTOs would be useful in the README?

PR #687, opened 2020-10-19
Feature/implement services

pedsf1968:
ISSUE #667 
ISSUE #691 
Disconnect Repositories from Controllers 
- Implement Data Object Transfert for each Entities to work with Controllers
- Implement Services that make connections between Repository Entitie and Controller DTO
- Implement unitary tests for new services
- Refactoring controllers tests
- Reorganise source tree directory


pedsf1968:
Hello @arey did you see my implementation of #667 #691 and refactoring?
I thing sources tree is better like this with the separation of Model, DTO, Services and Controllers.

dsyer:
Thanks for all the ideas. I would say this PR is too large really to consider as one change. It also goes against some decisions we made quite a long time ago (e.g. refactoring to vertical slices: 83ff9a50, and not using DTOs to keep the codebase as small as possible). So, while there might be some useful bits in there, it might be better to extract them and discuss first with the community.

PR #719, opened 2021-01-25
i18n + Features

aewinformatica:
* FIX Maven Project Build Lifecycle Mapping Problem
* update Owner list from better view Pets List
* messages - Include pt_BR
* Date Internationalization


pivotal-issuemaster:
@aewinformatica Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=719)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=719) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@aewinformatica Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-petclinic&pullRequestId=719)!

dsyer:
Some of this is interesting. I'm not sure we want the pom.xml change (we've discussed that before and concluded it should already be covered by the format plugin). The date format changes don't look like they would work ; and yyyy-MM-dd is an international date format anyway, so I'm not sure we need to change it. The net result of the i18n changes is that only English and Brazilian have a full set of messages? It might be better to leave the `pt_BR` messages in your fork and concentrate on better coverage for the languages that we already support.

PR #763, opened 2021-06-01
Upgrade to 2.5.0

Stef-33560:
Release Note 
 - upgrade Spring Boot to 2.5.0 and such other libs
 - add dependencies 
   - wro4j-extensions ; note sure about its utility
   - activeweb-lescc-maven-plugin : usefull, much recent libc compiler to build sucessfully SCSS
 - add missing mixins in src/main/less/petclinic.less
 - add stuff to build the whole package 
  -  pom.xml (build/test/package)
  - ci-settings.xml to define CI_JOB_TOKEN

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.m2/*
 target/*
 .settings/*
 .classpath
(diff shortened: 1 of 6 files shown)


dsyer:
For just the Spring Boot upgrade see 11f1234b. For the rest, maybe we can break it down into smaller chunks (not sure any of it is necessary).

PR #780, opened 2021-07-16
Add GitHub Action based build workflow

DerEros:


DerEros:
This wasn't meant to be sent here :)

PR #787, opened 2021-07-21
Create application-test.properties

Scythd:


dsyer:
Looks like a hard-coded database URL for a host that won't work for anyone else. So probably not a good idea for this project.

PR #791, opened 2021-07-25
Update pom.xml

aman-deep739:
Skip checkstyle


snicoll:
@aman-deep739 not sure if you actually intended to submit this as a proposal but we don't want to skip checkstyle.

PR #798, opened 2021-08-06
Added services keywords docker-compose.yml

kimkuma:
Hi 

I started the docker-compose file included in the existing project.

I added it because I thought something was missing.

please confirm

thanks. :)


dsyer:
It works for me with the file we have. Maybe it depends on the version of docker-compose that you used (I have 1.29.2)?

PR #809, opened 2021-09-08
Test task

BigPotatoo:


dsyer:
There's quite a lot in this patch. I'm not sure we are interested in large functional changes - the idea is to keep it as simple as possible. I'll leave it open in case anyone wants to discuss, and maybe outline what the changes are, and why they might be interesting.

Issue #91, opened 2015-07-01
Consider adding pagination to list views


eustachef:
Most real app will need pagination I guess, so it would be nice to have prototypical implementation.

liouxiao:
According to http://dandelion.github.io/components/datatables/1.1.0/docs/html/#4-2-feature-enablement, `dandelion-datatales` already supports pagination, and you'd turn on this feature by simply setting `pageable` attribute to true in related jsp files (e.g. `ownerList.jsp`), like this:

```
    <datatables:table id="owners" data="${selections}" row="owner" theme="bootstrap2" 
                      cssClass="table table-striped" pageable="true" info="false" export="pdf">
```

arey:
For large dataset (this in mostly the case in real application), server side pagination is the best. We may use dandelion pagination support. But DAO have also to be updated.

somayaj:
hi, can this be worked on?

simrin051:
Hi @arey @somayaj  

Can i take up this issue

arey:
@simrin051, I think so but we may ask confirmation to @snicoll or @dsyer

simrin051:
Hi @snicoll / @dsyer 

Can you please confirm whether i can take up this issue

simrin051:
Hi @arey

I have commented on the API logs issue. Can you please confirm.

dsyer:
I don't see the connection with the logs issue. If you want to make a proposal for pagination, go ahead. I think I would prefer to see something that didn't use additional libraries, but I guess we can assess that when we have something to look at. Pet Clinic has no Javascript in the web layer, and that does enable us to concentrate on the Spring side of things and not get caught up in JS dependency churn. Ideally we'd keep it that way IMO, but we can keep the options open while there is no code to review..

simrin051:
Thanks @dsyer  i will take up this issue

PR #812, opened 2021-09-13
Added Pagination to owners List and vet List

simrin051:
Added Pagination to owners List and vet List


simrin051:
Hi @dsyer 

Can you please merge the request if it is good to go ahead.

dsyer:
This is the right set of changes to make, consistent with my comments in #91. It adds extra parameters to repository anc controller methods without introducing new dependencies.

There are some issues with code format (you need to apply the spring-javaformat plugin).

I'm not sure about some of the details (e.g. you added a redundant `Model` parameter to one of the methods, and there is a new `@Service` layer that we haven't needed before).

Maybe we can tidy up the formatting and we could review the changes in more detail?

simrin051:
sure @dsyer  I will look and resolve by today

simrin051:
Hi @dsyer 

Can you please have a look. I have formatted the code and rebased it. Conflict is resolved.

simrin051:
Hi @dsyer 

Can you please review the source code.  I have formatted the code.
image
image

dsyer:
Code formatting is good now, thanks.

It looks like there are some changes in `Owner` (XML annotations and eager fetching). Was that necessary?

I'm getting test failures. Did you run the tests? Probably you forgot to create mocks for your new service layer? I'm not sure I like the service layer really. We didn't need it before, and I only see one reference to the `OwnerService` (for instance), so it seems like it could be a convenience method in the controller.

Also it looks like maybe you could look for things like unused instance variables and tidy that up (e.g. `OwnerController` has an unused `totalPages`). Your IDE will tell you where they are.

I don't like the formatting of the pagination links in the tables, but we can probably fix that later.

dsyer:
Here's my suggestion for the page controls (immediately below the table):

```html
  <div th:if="${totalPages > 1}">
    <span>Pages:</span>
    <span>[</span>
    <span th:each="i: ${#numbers.sequence(1, totalPages)}">
      <a th:if="${currentPage != i}" th:href="@{'/ownerpage/' + ${i}}">[[${i}]]</a>
      <span th:unless="${currentPage != i}">[[${i}]]</span>
    </span>
    <span>]&nbsp;</span>
    <span>
      <a th:if="${currentPage > 1}" th:href="@{'/ownerpage/1'}" title="First" class="glyphicon glyphicon-backward"></a>
      <span th:unless="${currentPage > 1}" title="First" class="glyphicon glyphicon-backward"></span>
    </span>
    <span>
      <a th:if="${currentPage > 1}" th:href="@{'/ownerpage/' + ${currentPage - 1}}" title="Previous"
        class="glyphicon glyphicon-triangle-left"></a>
      <span th:unless="${currentPage > 1}" title="Previous" class="glyphicon glyphicon-triangle-left"></span>
    </span>
    <span>
      <a th:if="${currentPage < totalPages}" th:href="@{'/ownerpage/' + ${currentPage + 1}}" title="Next"
        class="glyphicon glyphicon-triangle-right"></a>
      <span th:unless="${currentPage < totalPages}" title="Next" class="glyphicon glyphicon-triangle-right"></span>
    </span>
    <span>
      <a th:if="${currentPage < totalPages}" th:href="@{'/ownerpage/' + ${totalPages}}" title="Last"
        class="glyphicon glyphicon-forward"></a>
      <span th:unless="${currentPage < totalPages}" title="Last" class="glyphicon glyphicon-forward"></span>
    </span>
  </div>
```

simrin051:
I got the below error message hence i had to do eager fetching. Eager fetching solved the issue
Caused by: org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role :org.springframework.samples.petclinic.owner.Owner.pets, could not initialize proxy - no Session


I have removed service failure. Added the code in controllers

I have removed unused variables from my code

Pagination i have included your code. It looks great

simrin051:
Can you please review my code

dsyer:
That's looking much cleaner, thanks. There are still those changes in `Owner` that I didn't understand. Was that intentional? Also, if we go back to "owners" as the field name in `OwnerController` then the changes will be smaller.

The only thing left in my mind is the best path for the paginated resource. I think `/owners?page=1` seems better than `/ownerpages/1`, and then we could default the parameter value and we'd only have one endpoint again for owner lists.

simrin051:
Yes  @dsyer  the changes in Owner class was intentional as i was getting error on lazy initialization. Eager initialization solved the issue.

I have amended the code:  owners variable name instead of ownerRepository variable name in the OwnerController class.

@GetMapping("/owners")  is used for the method processFindForm. If i use  same end point for pagination it  does give error as one mapping for one method. Kindly please advice @dsyer

dsyer:
I don't see any changes. Did you forget to push?

You won't be able to use the same URL path for 2 methods, but surely only one method is needed for `@GetMapping("/owners")`?

CAn you be more specific about the lazy initialization errors? I don't think that explains the use of `@Xml*` annotations.

simrin051:
Hi @dsyer 

I have not pushed code yet.
I am sorry. I had not got your point earlier. I had forgotten that i used this Xml annotation. I have removed Xml annotations in owner class. 

Dyser can you please elaborate for the mapping. I am not clear.

dsyer:
I mean that I believe we ought to be able to get by with just one method for `/owners` with a request param for page number (defaulting to 1).

I also just noticed that `lastName` is an instance variable in `OwnerController`. That can't be right - it breaks the thread safety of the controller. Probably needs to be folded in as a method parameter.

simrin051:
Hi @dsyer 

yeah. I have not written test cases for these. Can you please confirm whether i am supposed to write test cases for it.

dsyer:
Yes, we need test cases. The ones taht were there before are currently broken. And you probably need to add tests for the pagination.

simrin051:
Sure Thanks i will do it

dsyer:
Great. It looks like we are getting close.

There are the tests that fail still. And a few issues to mop up in the `OwnerController`. Once we fix those we can do the same thing in the `VetController`. To see the issues in the controller have a look at my version above to see the differences:

* you call the repository twice to get one page of results where once is enough
* the `Pageable` parameter in the `@GetMapping` method is redundant (you only use the `pageNo` which is a request param)
* the extra `Model` is also redundant (there is already a model map)
* convenience methods can be private (no need for public)

simrin051:
@dsyer Thanks i have corrected the issues you have mentioned for Owner Controller  and Vet Controller.
After implementing pagination i used to get test cases failed  errors with the original test cases. Later i modified the test cases hence no test cases were failing.
Can you please  advice where the test cases failed with a screenshot if possible.

simrin051:
I found @dsyer 
Thanks. I am working on the test cases
 
image
image

simrin051:
Hi @dsyer 

Can you please review the code. I have made the test cases working.

dsyer:
Just looking at `VetControllerTests` for now, it looks like you removed the assertions from `testShowVetListHtml()` - the tests are green but they don't test any behaviour. It just sets up the mock repository and then doesn't use the `mockMvc`. I think we have the same problem with 2 or 3 tests in `OwnerControllerTests`.

Also you forgot to run `mvn spring-javaformat:apply` so the build fails with format errors.

simrin051:
@dsyer 

I was stuck in testShowVetListHtml()  for VetController it was showing page is Null error. I had to remove the assertion. Can you give me a clue on how to proceed.

dsyer:
In the case of the `VetControllerTests` I think you just need to fix the mock repository so it returns a `Page` instead of a `List`. Then you can go back to the test that we have on `main`. E.g.

```java
	@BeforeEach
	void setup() {
		...
		given(this.vets.findAll(Mockito.any())).willReturn(new PageImpl<>(Lists.newArrayList(james, helen)));
	}

	@Test
	void testShowVetListHtml() throws Exception {
		mockMvc.perform(get("/vets.html")).andExpect(status().isOk()).andExpect(model().attributeExists("vets"))
				.andExpect(view().name("vets/vetList"));
	}
```

dsyer:
That looks like it will work. You have duplicated the `given` with the `Pageable` though (we only need one).

simrin051:
@dsyer 

I have removed it. Can you please review the code.

dsyer:
Looks good. Green tests anyway. You need to remember to `mvn clean install` to check that the build is green before you push. You forgot (again) to `mvn spring-javaformat:apply`.

simrin051:
Thanks..
When i ran mvn clean install i got the below error.. Can you please advice. I noticed that 3.1.1 version of maven is used and i installed it but getting the below error.
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 8645 out of bounds for length 8192

  <groupId>org.apache.maven.plugins</groupId>
   <artifactId>maven-checkstyle-plugin</artifactId>
   <version>3.1.1</version>

dsyer:
You don’t need to install maven, just use the wrapper script (`mvnw`) in the root of the project.

simrin051:
Thanks @dsyer 
Build was successful. I remember that I had applied spring format plugin in the last push. I ran it in the root of the project. I have ran it in latest  push as well. Can you please let me know.

dsyer:
The build was still failing for me. Maybe you forgot to push a commit? Anyway, we are nearly there. We still need to make the change in `VetController` that we already did in `OwnerController` (to a single `@GetMapping` with a `@RequestParam` for the page number).

simrin051:
Thanks @dsyer 

I had changed Vet Controller with Request Param. If i am not wrong i had committed the file earlier. 
I have removed unused imports using IDE and removed * imports.

dsyer:
Sorry I think I was just confused because `VetController` has a private utility method that has a redundant `@PathVariable` parameter. I’ll check the rest and get back to you.

dsyer:
Merged with some polish in 8ad9c05

PR #813, opened 2021-09-16
Code Formatting

dos01:
Changed code formatting to read it easier.


dsyer:
We use the spring-javaformat plugin to normalize format including whitespace in this project. Thanks for the patch, but it doesn't conform with the plugin standard, so we'll have to decline.

PR #814, opened 2021-09-20
update db

paramjeet01:


dsyer:
This would be needed if you were running the app as a docker-compose service (for example). That isn't supported in the `docker-compose.yml` that we provide, so I don't think we should do this.

PR #821, opened 2021-09-24
Taller repository

javibanda:


dsyer:
This looks like it's probably spam, but there's quite a bit of work there. Maybe take it to https://github.com/spring-petclinic if you want to showcase a new stack?

PR #826, opened 2021-10-03
Switch CI/CD from Travis to GitHub Actions

arey:
The [travis build](https://travis-ci.org/spring-projects/spring-petclinic/) did not work since a few months.
I propose to switch to GitHub actions. You will check if PR are compiling and all tests are OK.
It's a default Maven workflow. Later we could add format check or wathever.


dsyer:
This seems like a good idea. Can we use the `mvnw` wrapper instead of the installed `mvn` from the container?

dsyer:
BTW @snicoll and I need to contact you out of band. What is the best email address to use? (We tried the one in your github profile, but there was no response.)

arey:
I've changed the worklow to use the maven wrapper. It looks like the cache is working.

My @free.fr email is the good one and I received the mail sent by @snicoll. It looks like he didn't received my answer. I've just resent it and added you as cc. Le me know.

PR #868, opened 2021-11-19
Convert less to scss

dsyer:
Modern versions of Bootstrap use scss and don't provide less source code any more. Wro4j doesn't have good support for the latest form of scss, so we have to use node.js to compile the petclinic.css. There were also a few tweaks to styles in menus and buttons because of migrations in Bootstrap. Finally the glyphicons (used in navbar) had to be replaced with font-awesome.


arey:
Good job Dave.
I'm disappointed that in the Java world we don't have a tool to generate `CSS` from `SCSS`. One more sign that the Java ecosystem is withdrawing from the frontend?
I find it a shame to have to go through `Node.js`. I find the process a bit heavy.
Moreover, in the past, behind a corporate proxy, I have already had problems with the use of the `frontend-maven-plugin` (node download blocked).
Are there not other solutions?

dsyer:
I don’t know to be honest, but I couldn’t find a way to make the wro4j thing work, and it seems like it would only work with jruby anyway. The future is JavaScript it seems. Maybe we could use GraalVM instead of the front end plugin, but I don’t know if that’s any more palatable?

dsyer:
Here's another idea: what about if we simply commit `petclinic.css` to source control, and add a profile that can be used to re-build it (using frontend plugin)? That way everyone gets something that works and there is a well-defined process for building if there are style changes (which are rare).

dsyer:
I found another Maven plugin. It looks mostly abandoned, and it depends on https://github.com/bit3/jsass which is definitely not maintained, but it was updated a year ago, and it works. See 0a2b19c.

arey:
I saw that many users open an issue because they don't find the `petclinic.css` file. Generate this file once through a maven profile and commit it could be a way to avoid this kind of mistake (due to IDE configuration I mean). It may be documented in the readme.

For the moment, I prefer the approach to use the deprecated `libsass-maven-plugin` instead of introducing `node` and `npm` installation. It's personal.

arey:
I saw that many users open an issue because they don't find the `petclinic.css` file. Generate this file once through a maven profile and commit it could be a way to avoid this kind of mistake (due to IDE configuration I mean). It may be documented in the readme.

For the moment, I prefer the approach to use the deprecated `libsass-maven-plugin` instead of introducing `node` and `npm` installation. It's personal.

dsyer:
OK. I put the CSS generation in a profile. If we retain the git history we will be able to see all the choices if ever they need to me revisited. I'm going to merge.

PR #873, opened 2021-11-25
i18n for Farsi.

aliazani:
Add message for Persian language.

### DIFF
--- a/src/main/resources/messages/messages_fa.properties
+++ b/src/main/resources/messages/messages_fa.properties
@@ -1,0 +1,8 @@
+welcome=خوش آمدید
+required=الزامی است
+notFound=پیدا نشد
+duplicate=در حال استفاده است
+nonNumeric=باید همه عددی باشد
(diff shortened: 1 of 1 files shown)


dsyer:
I guess this is interesting, but it's sort of academic right, because the app contains plenty of Strings that are not translated? We probably can't really ever claim that we aim to support all (or any) languages without doing some more fundamental work on extracting more text into messages.

However, if we were to start accepting new translations for the existing minimal set of labels, is there a particular reason that Farsi should be the first?

aliazani:
> I guess this is interesting, but it's sort of academic right, because the app contains plenty of Strings that are not translated? We probably can't really ever claim that we aim to support all (or any) languages without doing some more fundamental work on extracting more text into messages.

Yes, I think that's right about academic right things.
To answer your question about any reason that Farsi should be the first, Short answer is no.
There is no superiority for Farsi against German or any other languages.
I just try to contribute for the I18n for Farsi, the language that I know well as native person.

Issue #636, opened 2020-06-25
Add support for Postgres


dmadunic:
in addition to h2, Petclinic app comes with support for mysql, it would be nice to have similar support for the second (maybe first ;-)) most popular open source database Postgres.

Support can be done along the same lines as the one for mysql, namely to add the following files and folders:
+ resources/
    - application-postgres.properties
    + db/
        + postgres/
            - data.sql
           - schema.sql

And finally, posgres jdbc driver should be added to the runtime dependencies in pom.xml

I have also attached to this issue all of the above files, for review.

[resources.zip](https://github.com/spring-projects/spring-petclinic/files/4829935/resources.zip)
schema.sql

arey:
@dmadunic if you are interested, to prepare a Pull Request, you could look at the legacy version [spring-petclinic/spring-framework-petclinic](https://github.com/spring-petclinic/spring-framework-petclinic) that support PosgreSQL (see the [readme.md](https://github.com/spring-petclinic/spring-framework-petclinic/blob/master/readme.md))

arey:
Ok great. Before that wait a response from @dsyer or @snicoll

snicoll:
Feel free to submit a PR and we can take it from there.

nroi:
I'd like to revive this topic. It's been over a year since the last reply, so I hope it's ok that I just went ahead and submitted a PR.

lynchmaniac:
@nroi what a synchronicity, I just also posted a PR too on the same subject. :smiley: 

Cool, hoping the code is merged.

nroi:
@lynchmaniac aah, what are the chances, after more than a year of silence, two PRs are posted just 3 minutes apart :smile:

dsyer:
Resolved via #878

PR #878, opened 2021-12-07
Add support for PostgreSQL

nroi:
See #636

The following two lines in the `schema.sql` may need some explanation:
```sql
DROP SCHEMA public cascade;
CREATE SCHEMA public;
```

The bottom line is that those two lines ensure that we can start and restart the application as often as desired and insert new data through the web interface before restarting the application, without running into unique constraint violations or similar problems. But the data we have inserted through the web interface always disappears after a restart, unlike the MySQL version, where data remains available after an application restart.

The long explanation:

The goal is to provide an idempotent SQL file so that we can restart the application as often as desired. Without those two lines, the statements in `schema.sql` could not be executed after an application restart because the tables already exist. So the entire schema (including all tables, indexes etc.) is removed, and an empty schema is created.

The corresponding `schema.sql` file for MySQL is using a different approach: Instead of removing all tables and all data, each individual statement is idempotent (e.g. by using `CREATE TABLE IF NOT EXISTS` instead of just `CREATE TABLE`). With PostgreSQL, this would not work: We need to update the sequence values in the `data.sql` file (e.g. `ALTER SEQUENCE pets_id_seq RESTART WITH 14;`) because otherwise, Postgres would try to insert new values (e.g. when adding a new pet in the web interface) with a starting ID of 1 and cause a unique constraint violation, since a pet with an ID of `1` has already been inserted via `data.sql`. So we set the start value of the sequence to a literal value, like `14` for the `pets` table. However, this is based on the assumption that the table includes only the 13 pets that we have previously inserted with this script. So, this is not going to work if the table were to still include pets that have previously been inserted through the web interface.

I've also looked at how they have solved this problem at [spring-framework-petclinic](https://github.com/spring-petclinic/spring-framework-petclinic), they have simply used a larger starting value for the sequences, for example:

```sql
ALTER SEQUENCE visits_id_seq RESTART WITH 100;
```

Other than that, they have used the same approach as with MySQL (i.e., idempotent statements such as `CREATE TABLE IF NOT EXISTS`).

This works fine as long as you don't insert too many pets with the web interface. But if you insert 100 pets, and then restart the application, and then add another pet, you're going to have the unique constraint violation again.


dsyer:
LGTM. We can fix the sequence issue later, if anyone wants to contribute something.

PR #879, opened 2021-12-07
feat: add support for Postgres #636

lynchmaniac:
With this code, you can add support for Postgres to this project. It's related to issue #636

For example, you can run test this way

```sh
mvn test -Dspring.profiles.active=postgresql,spring-data-jpa
```

Or launch jar this way
```sh
java -jar -Dspring.profiles.active=postgresql,spring-data-jpa /app/app.jar
```
 
You must have a working Postgres installation.
Feel free to comments.


dsyer:
This looks good. Any reason you didn't follow the existing naming convention for `{schema,data}.sql`?

lynchmaniac:
These scripts is based on work of Vitaliy Fedoriv and Antoine Rey from [here](https://github.com/spring-projects/spring-petclinic/blob/main/src/main/resources/db/postgresql/petclinic_db_setup_postgresql.txt).

I didn't want to change their job. But there is no technical problem to rename the files.
Do you want me to rename the SQL files ?
initDB.sql into schema.sql
populateDB.sql into data.sql

src/main/resources/db/postgresql/populateDB.sql
```diff
@@ -0,0 +1,80 @@
...
+INSERT INTO visits VALUES (2, 8, '2011-03-04', 'rabies shot') ON CONFLICT DO NOTHING;
+INSERT INTO visits VALUES (3, 8, '2009-06-04', 'neutered') ON CONFLICT DO NOTHING;
+INSERT INTO visits VALUES (4, 7, '2008-09-04', 'spayed') ON CONFLICT DO NOTHING;
+
+--Resetting the sequence to ensure idempotence
+SELECT setval('visits_id_seq', (SELECT MAX(id) FROM visits));
+
+INSERT INTO users(username,password,enabled) VALUES ('admin','{noop}admin', true) ON CONFLICT DO NOTHING;
+
+INSERT INTO roles (username, role) VALUES ('admin', 'ROLE_OWNER_ADMIN') ON CONFLICT DO NOTHING;
```
nroi, line 75: What's the purpose of the `users` and `roles` table? The `data.sql` and `schema.sql` files for MySQL don't have those tables.

dsyer:
> Do you want me to rename the SQL files ?

Yes, please. We'll probably have to address @nroi 's question about the users/roles. I assume it's like `user.sql` in the MySQL scripts, but the password doesn't match what we have in the readme.

src/main/resources/db/postgresql/populateDB.sql
lynchmaniac: it come from the work of Vitaliy Fedoriv and Antoine Rey from [here](https://github.com/spring-projects/spring-petclinic/blob/main/src/main/resources/db/postgresql/petclinic_db_setup_postgresql.txt).

They are unnecessary here in this project. I can take it off if you want. I just tested and it works perfectly without.

dsyer: Sounds good then.

lynchmaniac:
I push the sql's modification. Tell me if it's good for you or not :smiley:

dsyer:
I went ahead and merged #878 since it technically was first in the queue and didn't have any issues to resolve. Thanks for your help, everyone.

PR #883, opened 2021-12-15
feature#1: add tests and age column

midorun:


dsyer:
I don't think we need to change the entities. Also we don't need a Jenkinsfile and a bunch of other stuff in here. Thanks for the patch though, and happy holidays.

PR #895, opened 2022-01-04
Avoid dropping Postgres schema

ssapra:
The database user needs to be an owner of the public schema or a superuser in order to execute the schema.sql for database initialization (to drop the schema).

To avoid that assumption on privileges, we can instead re-create the table if it already exists.

This unblocks testing scenarios where the Postgres database user is not an owner of the public schema.

Authored-by: Shaan Sapra <shsapra@vmware.com>


dsyer:
This looks sensible. Is there no equivalent to the MySQL `create table if not exists`?

ssapra:
There is an equivalent in Postgres, but it would require more changes to the insert statements to avoid conflicts. MySQL addresses this with `INSERT IGNORE`. Postgres supports something similar with `ON CONFLICT DO NOTHING`. 

Would it make more sense to match the pattern established by `mysql` or follow `h2` and `hsqldb` which also drop the tables to simplify the data insertion?

dsyer:
IMO it would be better to support a permanent database with no destruction of the tables on startup (like MySql). What do you think the best option is?

ssapra:
Yes, I agree, it's better to be non-destructive so that an application restart will not affect existing data.

I've update the PR with an updated set of schema.sql and data.sql files.

PR #896, opened 2022-01-05
Remove PetRepository and use Owner as aggregate

dsyer:
Owner is really the aggregate root in DDD terms and there is no need to directly access the Pet entity.


dsyer:
Reviewed by @odrotbohm with :+1:

PR #922, opened 2022-02-22
Set up CI with Azure Pipelines

StephCol:
[skip ci]


dsyer:
We already use Github Actions. Is there a reason to switch?

dsyer:
Looks  like possibly spam.

PR #949, opened 2022-04-07
Ensure pets and visits are retained when owner edited 

### LINKED ISSUE #947: Changing owner information makes their pets disappear
When we change some information of an owner, their pets disappear from the PetClinic app. Tested on Windows 10, Java version 14.0.2, with both the integrated H2 database and MySQL.

Steps to reproduce:
Steps to reproduce:
-connect to the app -> Find owners -> Find owner
-George Franklin -> Edit owner
-set a new first name for the owner
-after clicking Update, the app shows the owner information page without the pets

tpsst5:
Resolves #947

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
@@ -92,6 +92,8 @@ public class Owner extends Person {
 		return this.pets;
 	}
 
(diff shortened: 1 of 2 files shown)


dsyer:
I assume this is spam since it causes a test to fail.

dsyer:
Thanks for the patch. I think we prefer to rely on the JPA model to be updated accurately (which requires the changes in d381fb658).

PR #962, opened 2022-04-26
Automatic code format with moderne.io and openrewrite recipe

arey:
Created with: [Moderne](https://app.moderne.io)


arey:
Not compatible with spring-javaformat

PR #981, opened 2022-06-04
5656Update PetType.java

alex-f8:


mdeinum:
Looks like spam and inserting control chars into the code. Is spam a thing nowadays on GitHub?

PR #990, opened 2022-06-15
Update javadoc: it incorrectly said to return 'true'

amaembo:


dsyer:
Merged in 2a12048 thanks.

PR #991, opened 2022-06-15
dev container tweaks

burkeholland:
Fixed location of Dockerfile and moved customizations out of features

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
 	"name": "Petclinic",
-	"dockerFile": "docker.languageserver.formatter.ignoreMultilineInstructions",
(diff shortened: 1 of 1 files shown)


dsyer:
Hmm. Sorry about the dockerfile error (I fixed that already on main and force pushed). I'm not so sure about the other changes. In particular "customizations" is part of "features" according to the editor in vscode. Did you actually see it work in the level above that?

dsyer:
I pushed another change that moves the extensions up another level. Seems to work for me. Maybe there are multiple versions of VSCode/Devcontainer that are not compatible with each other? Also I had to disable gradle to get the classpath to resolve correctly. Not sure why that is, but it seems you have to choose.

dsyer:
I think these changes are obsolete now. Thanks for the suggestions.

PR #997, opened 2022-06-18
Fixed incorrect package path

daejoon:


dsyer:
That other one is "deprecated since 2.7.0 for removal in 3.0.0" (per the java docs). I think we are fine.

PR #1034, opened 2022-08-22
Adding dockerfile

shivampateriya1:
I created a multistage dockerfile. It is used in 3 ways.
1. Git action (where we can add a step to build the whole application using docker file rather than building the application user don't need to download dependence like maven and java ) 
2. Docker-compose (we can up the whole application using docker-compose)
3. It will be helpful while creating k8s deployment 

Further tasks :
1. I will add the docker-compose file to up the application in one go with MySQL and Postgres.
2. I will also create k8s deployment for the same. It will be helpful to implement SRE and DevOps practices for the new learner.

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -1,0 +1,16 @@
+#multistage docker file 
+FROM openjdk:11 as builder
+
+WORKDIR /springapp
+
+COPY . /springapp/
+
+RUN ./mvnw -B package
+
+
+FROM openjdk:11 as petclinic 
+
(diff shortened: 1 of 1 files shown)


dsyer:
I think we would prefer to persuade to use the Spring Boot tooling to build a container image. That's a good encapsulation of best practice and takes the guess work out of writing a `Dockerfile`, so definitely a good idea for learners to get to grips with, and it's already documented in the README.

PR #1038, opened 2022-09-01
Upgrade to Spring Boot 2.7.3 and Gradle 7.5.1

dgcd:


pom.xml
```diff
@@ -3,12 +3,12 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.springframework.samples</groupId>
   <artifactId>spring-petclinic</artifactId>
-  <version>2.7.0-SNAPSHOT</version>
+  <version>2.7.3-SNAPSHOT</version>
```
dsyer, line 6: 2.7.3 is released, right? So no need for snapshots.

build.gradle
```diff
@@ -36,6 +36,10 @@ dependencies {
   testImplementation 'org.springframework.boot:spring-boot-starter-test'
 }
 
-test {
+tasks.named('test') {
   useJUnitPlatform()
 }
+
+tasks.named('jar') {
```
dsyer, line 43: I don't see this in start.spring.io. What's the rrationale?

dsyer (commented):
Thanks. Some comments inline.

pom.xml
dgcd: Fixed.

build.gradle
dgcd: By default it generates plain-jar alongside uber-jar.
It is not a problem, just unnecessary task and file.

dsyer:
Merged as 276880e. Thanks.

PR #1055, opened 2022-09-22
Readability improvements & Use stream instread of forEach

ckyeon:
## Commit Description
- 3720069
  - use stream instead of forEach
- de0f50d
  - Remove else statement
  - Remove unused method parameters
  - Arrange imports
---
If there is a problem with the change, please explain.
I want to contribute to this project!

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetTypeFormatter.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetTypeFormatter.java
@@ -51,8 +51,12 @@ public class PetTypeFormatter implements Formatter<PetType> {
 	@Override
 	public PetType parse(String text, Locale locale) throws ParseException {
 		Collection<PetType> findPetTypes = this.owners.findPetTypes();
-		return findPetTypes.stream().filter(petType -> petType.getName().equals(text)).findFirst()
-				.orElseThrow(() -> new ParseException("type not found: " + text, 0));
+		for (PetType type : findPetTypes) {
+			if (type.getName().equals(text)) {
+				return type;
+			}
+		}
+		throw new ParseException("type not found: " + text, 0);
 	}
 
 }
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/samples/petclinic/owner/PetTypeFormatter.java
```diff
@@ -51,12 +51,8 @@ public String print(PetType petType, Locale locale) {
 	@Override
 	public PetType parse(String text, Locale locale) throws ParseException {
 		Collection<PetType> findPetTypes = this.owners.findPetTypes();
-		for (PetType type : findPetTypes) {
-			if (type.getName().equals(text)) {
-				return type;
-			}
-		}
-		throw new ParseException("type not found: " + text, 0);
+		return findPetTypes.stream().filter(petType -> petType.getName().equals(text)).findFirst()
```
dsyer, line 54: I'm not sure this is an improvement really.

dsyer (changes requested):
Mostly pretty sane. I might prefer the else branches, but don't care much. Comments inline.

src/main/java/org/springframework/samples/petclinic/owner/PetTypeFormatter.java
ckyeon: It is a spring sample project, and it may not make sense because there is no performance improvement.

However, I thought it would be better for people who come to see the Spring sample project to see the Java 8 stream API than forEach.

I'm curious what you think.

---
It might be awkward because the sentence is being translated.
Please excuse me. 

I think I'm very lucky to talk to a good developer.
Thank you!

dsyer: I don’t think the purpose of this sample project is to promote the use of Java 8 APIs. It’s better to use whatever makes sense imo, and in this case the old syntax is shorter and easier to understand.

ckyeon: understand! 
commit 3720069 will roll back.

PR #1117, opened 2022-11-27
Install maven in Dev Container

harsha-s:
The documentation mentions maven but is not included in the dev container

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,8 @@
 ARG VARIANT=17-bullseye
 FROM mcr.microsoft.com/vscode/devcontainers/java:0-${VARIANT}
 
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,6 +1,8 @@
 ARG VARIANT=17-bullseye
 FROM mcr.microsoft.com/vscode/devcontainers/java:0-${VARIANT}
 
+RUN sudo apt update && sudo apt install -y maven && sudo apt-get clean
```
andrzejsydor, line 4: Is it required?

harsha-s: Yes, it was needed for my development work

dsyer: It is not required if you follow the instructions in the readme - the Maven wrapper is installed in the source code.

PR #1126, opened 2022-12-17
Dockerfile

Abdulra23:


andrzejsydor:
The commit message should be more descriptive.

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
+RUN apt-get update
+RUN mkdir test
```
andrzejsydor, line 3: Why `test` directory?

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
```
andrzejsydor, line 1: Why do you use ubuntu image just for a run jar?

andrzejsydor (changes requested):

mdeinum:
Spam. Please don't use a public repository for your tests!

PR #1133, opened 2022-12-28
add Dockerfile & chart directory

choni824:


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# add project
+ADD . /app/
+
+RUN gradle clean bootJar
+
+FROM docker.io/library/eclipse-temurin:11-jre-focal AS app
+
+WORKDIR /app
+COPY --from=build /app/build/libs/spring-petclinic-2.7.3.jar .
```
andrzejsydor, line 14: In my opinion, it would be better not to specify the version.
```suggestion
COPY --from=build /app/build/libs/spring-petclinic-*.jar spring.petclinic.jar
```

andrzejsydor (changes requested):

build.gradle
```diff
@@ -1,13 +1,13 @@
 plugins {
-  id 'org.springframework.boot' version '3.0.0'
+  id 'org.springframework.boot' version '2.7.1'
```
andrzejsydor, line 2: why not 2.7.7.

```suggestion
  id 'org.springframework.boot' version '2.7.7'
```

https://mvnrepository.com/artifact/org.springframework.boot/spring-boot

andrzejsydor (changes requested):

PR #1134, opened 2023-01-02
Using streams

IfeSunmola:
Refactored some methods from using foreach loops to streams

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
@@ -125,17 +125,12 @@ public class Owner extends Person {
 	 * @return a pet if pet name is already in use
 	 */
 	public Pet getPet(String name, boolean ignoreNew) {
-		name = name.toLowerCase();
-		for (Pet pet : getPets()) {
-			if (!ignoreNew || !pet.isNew()) {
-				String compName = pet.getName();
-				compName = compName == null ? "" : compName.toLowerCase();
-				if (compName.equals(name)) {
-					return pet;
-				}
-			}
-		}
-		return null;
+		final String tempName = name.toLowerCase();
+		return getPets().stream()
+				.filter(pet -> !ignoreNew || !pet.isNew())
+				.filter(pet -> pet.getName() != null && pet.getName().toLowerCase().equals(tempName))
+				.findFirst()
+				.orElse(null);
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/samples/petclinic/owner/Owner.java
```diff
@@ -130,17 +125,12 @@ public Pet getPet(Integer id) {
...
-				if (compName.equals(name)) {
-					return pet;
-				}
-			}
-		}
-		return null;
+		final String tempName = name.toLowerCase();
+		return getPets().stream()
+				.filter(pet -> !ignoreNew || !pet.isNew())
+				.filter(pet -> pet.getName() != null && pet.getName().toLowerCase().equals(tempName))
```
andrzejsydor, line 131: you can use `equalsIgnoreCase`

src/main/java/org/springframework/samples/petclinic/owner/PetTypeFormatter.java
```diff
@@ -50,13 +50,11 @@ public String print(PetType petType, Locale locale) {
...
 	@Override
 	public PetType parse(String text, Locale locale) throws ParseException {
-		Collection<PetType> findPetTypes = this.owners.findPetTypes();
-		for (PetType type : findPetTypes) {
-			if (type.getName().equals(text)) {
-				return type;
-			}
-		}
-		throw new ParseException("type not found: " + text, 0);
+		return owners.findPetTypes()
```
andrzejsydor, line 53:
```suggestion
		return owners.findPetTypes().stream()
```

andrzejsydor (changes requested):

dsyer:
I'm not sure the stream API is any better for these tiny examples, and might even be worse - it just looks "clever" to me, not really self-documenting or anything useful. I'm tempted to just close this PR, but interested to hear other opinions.

IfeSunmola:
> I'm not sure the stream API is any better for these tiny examples, and might even be worse - it just looks "clever" to me, not really self-documenting or anything useful. I'm tempted to just close this PR, but interested to hear other opinions.

Hi, 

I'm assuming this project is somewhat directed towards people that are new to java and spring? If that's the case, isn't it better for them to know that streams exists, and see the usage? Anyone that has made it here probably knows how to use the good ol for loop. 

Streams have been around for a relatively long time and questions about it even comes up in interviews. I didn't know about them till like a year ago because almost every code on the internet is still stuck on for loops

dsyer:
And that’s kind of my point right there. The purpose of this project is definitely to educate, but the subject is Spring and Java, not the JDK. If for loops work and are understandable by a larger audience, that’s a good reason not to change anything. There really is no upside.

PR #1135, opened 2023-01-02
Resource bundle not included in native image

DuncanCasteleyn:
The resource bundle was not included in the native hints which would cause the text to not be displayed in the right language.

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/PetClinicRuntimeHints.java
+++ b/src/main/java/org/springframework/samples/petclinic/PetClinicRuntimeHints.java
@@ -24,6 +24,7 @@ public class PetClinicRuntimeHints implements RuntimeHintsRegistrar {
 	@Override
 	public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
 		hints.resources().registerPattern("db/*"); // https://github.com/spring-projects/spring-boot/issues/32654
+		hints.resources().registerPattern("messages/*");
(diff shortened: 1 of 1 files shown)


dsyer:
The messages resource is a weird one - I believe it works if you remove devtools, so it's really a problem in Spring Boot, and may already be fixed.

patbaumgartner:
The resource bundle  issue is still in Spring Boot 3.0.1 - Looking forward to testing it with 3.0.2

dsyer:
Follow this for updates: https://github.com/spring-projects/spring-boot/issues/32853

sdeleuze:
Unless I miss something, this resource bundle is not included because by design Boot include only resources with recognized path and `messages/messages` is a custom resource bundle from the application. I had to include a similar hint [here](https://github.com/sdeleuze/spring-petclinic-data-jdbc/blob/6a079e054ea46f2fd05ef5b5cd84ee2b40594b14/src/main/java/org/springframework/samples/petclinic/PetClinicApplication.java#L50C) as well (my version does not use Devtools).

patbaumgartner:
I added a PR #1157 to fix native image generation because the native plugin is also missing. Even start.spring.io adds it to the generated sources.

dsyer:
I'm happy to merge this if the conflict is resolved.

dsyer:
Closed in cc46671

PR #1140, opened 2023-01-11
Try and use Spring native

anthonydahanne:


dsyer:
This should all work out of the box (with the "native" profile in Maven, and with the corresponding task in Gradle). The messages resource is a weird one - I believe it works if you remove devtools, so it's really a problem in Spring Boot, and may already be fixed.

PR #1150, opened 2023-01-22
Create .gitattributes

### LINKED ISSUE #1149: Add .gitattributes for mvnw file
In Git for Windows, the default newline is `CRLF`. But the `mvnw` file can only be interpreted by `LF` newline character.

I think it should add an `.gitattributes` file with the following content:

```txt
mvnw    text eol=lf
```

doggy8088:
fixes #1149

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 mvnw    text eol=lf
+*.java  text eol=lf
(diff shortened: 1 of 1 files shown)


dsyer:
IMO this should be the default for all local development (hence the user's responsibility), and not just for `*.java`. Happy to hear counter arguments.

PR #1157, opened 2023-02-06
Updating Project with the newest wrappers, dependencies, and docker containers

patbaumgartner:
Fixing native image generation

- Updating maven and gradle wrapper
- Upgrading project to Spring Boot 3.0.2 and dependencies/plugins
- Adding hint for resource bundles -> message.properties (not solved just by upgrading)
- Adding native image plugin because of PARSE ERROR and NonJtaPlattform Exception (known issue which is produced by Hikari)


patbaumgartner:
Hello @dsyer and @snicoll 

I think, I have fixed known issues for native images in the Petclinic. Please tell me what is missing for getting my changes approved and merged?

TIA,
Patrick

src/main/java/org/springframework/samples/petclinic/PetClinicRuntimeHints.java
```diff
@@ -24,6 +24,8 @@ public class PetClinicRuntimeHints implements RuntimeHintsRegistrar {
 	@Override
 	public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
 		hints.resources().registerPattern("db/*"); // https://github.com/spring-projects/spring-boot/issues/32654
+		hints.resources().registerPattern("messages/*");
```
dsyer, line 27: I don't think any changes to this file should be needed. Seems to work for me anyway.

pom.xml
```diff
@@ -20,12 +20,12 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
     <!-- Web dependencies -->
-    <webjars-bootstrap.version>5.1.3</webjars-bootstrap.version>
+    <webjars-bootstrap.version>5.2.3</webjars-bootstrap.version>
```
dsyer, line 23: If you change this you should also change the templates where it is loaded in HTML (and probably regenerate the css as well?). Maybe leave this change for another PR?

dsyer:
Rebased in 0962ed7

PR #1164, opened 2023-02-16
Java 19

ratulb:
Port to java 19.


Dockerfile
```diff
@@ -0,0 +1,13 @@
+# syntax=docker/dockerfile:1
+
+FROM openjdk:21-buster
```
andrzejsydor, line 3: Shouldn't be version 19, to be consistent with other things.

ratulb: Set base image to openjdk:19-buster from openjdk:21-buster to make sure it is consistent with rest of the changes. Tested locally.

dsyer:
No thanks, I think we'll stick with 17 (LTS). And we don't need a `Dockerfile`.

PR #1218, opened 2023-04-06
Dashboard

tenzinM:
Number of Owners in a selected City
Number of cities the Database


dsyer:
It's kind of an interesting change, and it looks like it might work. We don't really need any new features though, unless you can explain a bit more what motivates it.

PR #1231, opened 2023-04-18
Update welcome.html

EoinGalvin:


andrzejsydor (changes requested):
What is the PR about?
Not needed, to remove it.

PR #1234, opened 2023-04-22
Owener class's addVisit() method return type change to void

jcw1031:
Unused return value


xerx593:
it is unused, but it could (save 1 line):
- In [VisitContoller](/spring-projects/spring-petclinic/tree/main/src/main/java/org/springframework/samples/petclinic/owner/VisitController.java#L89-L90)
- and [test](/spring-projects/spring-petclinic/tree/main/src/test/java/org/springframework/samples/petclinic/service/ClinicServiceTests.java#L205-L206)

So it is rather a question of:
- "fluent api design" or not?
- how consequently (other objects?) and deep (all setters?)?

dsyer:
I'm not sure I care much one way or the other. Probably we can merge this - saving a line of code seems like a small gain.

PR #1236, opened 2023-04-25
Fix crashcontroller test

xerx593:
It is hard to summarize the where/why/when, but: `@WebMvcTest` is not (comparable with) a "full blown" (embedded) spring-boot application... (And i fear, we will have to wait a while, till that happens)

So I decided, to introduce:
- a "very slim" unit test
- and a (slim as possible, "full blown") integration test


dsyer:
Squashed and merged 2249cbc

PR #1238, opened 2023-04-29
Adding support for Azure Spring Apps with PostgreSQL

nevenc:
Added support for deploying spring-petclinic to Azure Spring Apps using PostgreSQL.


jaididon:
How to use file in intellij idea for Java spring boot??

nevenc:
> How to use file in intellij idea for Java spring boot??

What do you mean @jaididon?
What file specifically?
What are you trying to do?

jaididon:
File Specially
index.java

jaididon:
I am trying to do coding of Java springboot
But I don't have a clue how to create index.java in intellij idea

dsyer:
People should be able to use Petclinic on Azure without changing the source code. Maybe instead of a new properties file we could just add the comment about the JAVA_OPTS in the existing one? Or (better) in the txt file that lives near the SQL scripts?

PR #1253, opened 2023-05-22
Upgrade for Spring Boot 3.1

patbaumgartner:
Upgrade to Spring Boot 3.1
- Updating Wrappers (Maven and Gradle)
- Removing Spring Java Format Config: Current default is at Java 11
- Updating PostgreSQL Docker Image
- Updating Plugin Version
- Adjusting Build-Info since `maven.compiler.source` and `maven.compiler.target` got removed from Spring Boot Parent because `maven.compiler.release` is already set.  #gh-34365
- Replacing by Pivotal with by VMware Tanzu (SVG).


dsyer:
Could we extract the SVG into a separate file? It looks kind of ugly in line like that IMO.

dsyer:
The spring.io website is a React.js app (I think) so any styles and logos that appear in thesite might be pre-processed during the build. I'm not sure I understand your suggestion to "add the styles into the SVG".

dsyer:
You didn't want to use `th:src="@{...` ?

patbaumgartner:
Well spotted @dsyer - Thank you for recognizing. I made it like all the other images.

PR #1259, opened 2023-06-06
set digit length to 10

arsh1598:


dsyer:
Doesn't that duplicate the existing `@Digits` annotation? The intent seems to be the same at least.

arsh1598:
> Doesn't that duplicate the existing `@Digits` annotation? The intent seems to be the same at least.

@digits sets maximum digit limit not minimum.

src/main/java/org/springframework/samples/petclinic/owner/Owner.java
```diff
@@ -56,7 +56,7 @@ public class Owner extends Person {
 
 	@Column(name = "telephone")
 	@NotEmpty
```
andrzejsydor, line 58: so the not empty is probably redundant?

dsyer: Don't know. What would be the effect of removing it? None of these errors has any error messages or nice handling associated with them anyway, so it's all kind of moot.

arsh1598:
> I see. But not all telephone numbers have 10 digits, so `@Digits` seems like the right tool to me still. It's not really a big deal.

closing this PR

PR #1271, opened 2023-06-26
Refactoring Owner getPets function to use lambda functions

RafaelFantinel:


dsyer:
We've had similar proposals in the past and decided not to merge them. This one has limited scope so it is a good pull request. Not sure we really need it though, still.

RafaelFantinel:
I understand, I see it as a good modification to discourage the use of the traditional "for" and encourage the use of lambdas functions, bringing greater clarity and evolution in code writing.

PR #1274, opened 2023-06-27
Fix typo in VetController.java

jee14:
Erasing unnecessary 1 line.


andrzejsydor (changes requested):
rejected, there is no typo fixed

dsyer:
Kind of super picky, even by my standards. But that's fine.

PR #1278, opened 2023-06-30
Added Validation for Owner_ID and Pet_Id on PetController

4313testing:


dsyer:
It's a bigger change than I'd like really, but I'll go ahead and merge it if you can add tests (and fix the ones that you changed).

dsyer:
I don't see a need for all this code really. Can we just do some checking and throw `IllegalArgumentException`?

PR #1288, opened 2023-07-11
adding pet name validation on edit Pet

4313testing:
fixing duplicate pet name issue when editing a pet


dsyer:
I rolled this back because the build failed.

PR #1293, opened 2023-07-13
Handling Null Pointer Exception for Owner ID

4313testing:
handling null pointer Exception on Owner ID with IllegalArgumentException

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
@@ -57,7 +57,6 @@ class PetController {
 	@ModelAttribute("owner")
 	public Owner findOwner(@PathVariable("ownerId") int ownerId) {
 
-
 		Owner owner = this.owners.findById(ownerId);
 		if (owner == null) {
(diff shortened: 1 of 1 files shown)


dsyer:
The result is the same for this and the code we already have. The only difference is the exception type, which doesn't show up in the web UI anyway, unless someone switches on error attributes includes manually. The logs in the server are slightly more verbose (messy) with the `DataAccessException`, but I think we'll live with it.

Also, the check for `<=0` was redundant.

4313testing:
it was your request to use IllegalArgumentException for the Null Pointer so i just did what you suggested

dsyer:
Sorry, I think I closed the wrong one. I'll reopen and can you remove the `<=0` check since it isn't adding anything? Please rebase and squash into one commit.

4313testing:
okay i will do that

4313testing:
@dsyer sorry i had some conflicts in this branch 
i created a new PR #1294 
please check it

dsyer:
I think you muffed it up somehow - I can see 4 commits and 0 files changed. I was expecting 1 commit and 1 file changed.

PR #1295, opened 2023-07-16
Modified files:

CSS-MB:
.devcontainer/devcontainer.json
```


dsyer:
This is a bad idea really. You have to choose either Maven or Gradle or they get in the way of each other. It's an arbitrary choice, but it doesn't affect the user experience much.

PR #1298, opened 2023-07-21
Add Qodana configuration

saimirgasa:


dsyer:
Thanks for the patch. We probably have enough static code analyzers installed already for now.

PR #1303, opened 2023-07-24
Fix #1302 n+1 queries in owners

### LINKED ISSUE #1302: N+1 Queries in Owners endpoint
While digging around the Spring Petclinic I found this issue running [digma.ai](http://digma.ai/) on the project. The n+1 situation arises in two situations:

1) Making a parameterless GET request on /owners
2) Using the webpage and searching for an empty string

We can see that in the traces, side panel, and even the query behind the issue. This issue can be replicated using a simple shell script to concurrently load the /owners endpoint or even just running an empty string search a few times from the webpage.
BeforeTrace
SidePaneView
AfflictingQuery

My proposed fix resolves the n+1 issue and reduces the amount of un-necessary queries. We can see that in the new traces and the updated side panel view when testing in the same manner. Here's the new query I used to resolve this issue.
AfterTrace
AfterOverview
AfterQuery

tekshteint:
Issue #1302 
When using the parameterless GET request on the /owners endpoint as well as searching for an empty string on the webpage for owners, the petclinic application ends up creating N+1 queries. This is due to the following code:

OwnerController.processFindForm
```
                if (owner.getLastName() == null) {
			owner.setLastName(""); // empty string signifies broadest possible search
		}

		
		Page<Owner> ownersResults = findPaginatedForOwnersLastName(page, owner.getLastName());
		if (ownersResults.isEmpty()) {
			// no owners found
			result.rejectValue("lastName", "notFound", "not found");
			return "owners/findOwners";
		}
```

This can be solved by adding a new query to the Owner Repository and updating the controller. These changes can be observed using [digma.ai](http://digma.ai/) as I've left the necessary imports/dependencies for Open Telemetry as well.


dsyer:
It's not a terrible idea, but maybe not really all that necessary (given that n=5). Anyway, if you would like to remove the otel stuff and squash and rebase to a single commit I will take another look.

andrzejsydor (approved):

tekshteint:
> 

Not a problem, sorry for the mess as I'm still new to Git but I hope that this should be more proper now.

Branches squashed and merged, OTEL has been removed and only the fix remains.

dsyer:
The changes to `build.gradle` are just whitespace right? So we don't need them. Also the build failed (error message tells you how to fix it).

tekshteint:
@dsyer Everything should be taken care of, thanks again for your patience with me.

andrzejsydor (changes requested):
the commit message is different that the changes

andrzejsydor (approved):

src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
```diff
@@ -34,6 +35,7 @@
 import org.springframework.web.servlet.ModelAndView;
 
 import jakarta.validation.Valid;
+import org.thymeleaf.util.StringUtils;
```
dsyer, line 38: Please use Spring's `StringUtils`.

src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
```diff
@@ -112,11 +124,10 @@ public String processFindForm(@RequestParam(defaultValue = "1") int page, Owner
 
 	private String addPaginationModel(int page, Model model, Page<Owner> paginated) {
 		model.addAttribute("listOwners", paginated);
-		List<Owner> listOwners = paginated.getContent();
 		model.addAttribute("currentPage", page);
 		model.addAttribute("totalPages", paginated.getTotalPages());
 		model.addAttribute("totalItems", paginated.getTotalElements());
-		model.addAttribute("listOwners", listOwners);
+		model.addAttribute("listOwners", paginated);
```
dsyer, line 130: "listOwners" is now added twice?

build.gradle
```diff
@@ -1,8 +1,8 @@
 plugins {
-  id 'java'
-  id 'org.springframework.boot' version '3.1.1'
-  id 'io.spring.dependency-management' version '1.1.0'
-  id 'org.graalvm.buildtools.native' version '0.9.23'
+    id 'java'
```
dsyer, line 2: Please get rid of whitespace changes (I don't think this file changed at all so you could just grab it from main).

dsyer (changes requested):
Some more minor stuff to tweak.

PR #1311, opened 2023-07-29
compatibility with oracle database included

gut0leao:
Hi, Folks.

I've added compatibility with oracle database trying to change lesss prossible of the original code . I've tried to maintain all files in current patterns.

Cheers!


dsyer:
I'm not sure we want yet another jar file in the classpath. But I'm happy to merge it and put it in a branch so others can see what is needed. Could you please revert the changes to the Dockerfile and docker-compose.yaml (and add a new service profile for oracle in the same pattern as we already have if that makes sense)?

dsyer:
#1410 is a more complete patch

PR #1317, opened 2023-08-09
Change validation annotations and whitespace handling

evga7:
Replaced the @NotEmpty annotation with @NotBlank for improved validation accuracy, ensuring that the field is not only present but also not consisting of only whitespaces.

Additionally, updated the validation checks to use StringUtils.hasText instead of StringUtils.hasLength to account for whitespace-only values as well.


dsyer:
I guess I'm not totally opposed to this, and it's quite a small change. Could you fix the test though, please?

evga7:
I fixed it. Please check. thank you.

PR #1329, opened 2023-08-27
Changes to application.properties file to open h2-console in browser. 

soydivision:
Following README  was trying to access h2 console using _http://localhost:8080/h2-console_ . The state of the code at the moment doesn't allow to do so (at least I haven't found out the way). Adding two lines of prop did the trick.


dsyer:
Makes sense. I think the default for the enabled flag changed.

PR #1346, opened 2023-09-12
Update findOwners.html

Bobzero23:
I removed the 'add owner' button which was supposed to be in the navbar and not in the page body


dsyer:
No CLA. Probably not a serious request?

PR #1347, opened 2023-09-12
Update layout.html

Bobzero23:
I have added the 'add owner' to the navbar but make sure to change the behavior of the 'add owner' in this file.. make it work as you want (e.g where to go when you click the button and so on)


Alireza9476:
I would place the new menu after ,,FIND OWNERS", or at least before the ,,ERROR" menu, otherwise seems good.

dsyer:
Thanks, but probably we don't need it, and it seems a) incomplete and b) lacking a CLA.

PR #1392, opened 2023-11-06
adding Dockerfile

xxsotoxx:


dsyer:
Thanks, but we prefer to use a buildpack (`mvn spring-boot:build-image` is already available).

PR #1397, opened 2023-11-15
Checkstyle NoHttp suppressed for all log files

roxspring:
`NoHttp` rule should presumably be focused on files that might be committed in to the repository. This change expands an existing rule suppression to cover all `.log` files rather than just one specific `build.log`. This avoids log files written by external tools from failing `./mvnw test` execution.


roxspring:
@pivotal-cla This is an Obvious Fix

PR #1398, opened 2023-11-16
Add Dockerfile

amitbb92:
Dockerfile added with Project Built Locally


dsyer:
Thanks, but we prefer to use a buildpack (`mvn spring-boot:build-image` is already available).

PR #1399, opened 2023-11-18
Update dependencies

patbaumgartner:
- Updated Spring Boot to 3.1.5
- Updated Maven and Gradle Wrappers
- Updated Plugin and Dependencies
- Update Docker-Compose containers

- Removed Testcontainers Dependency Management due it's duplication from the spring-boot-parent


patbaumgartner:
It fails because of a to new version of docker which is incompatible with Spring Boot 3.1 
`Caused by: java.lang.NullPointerException: Cannot invoke "org.springframework.boot.docker.compose.core.DockerCliInspectResponse.hostConfig()" because "inspectResponse" is null` 

This will be solved in 3.2 -> https://github.com/spring-projects/spring-boot/issues/37991

dsyer:
Obsolete, I guess. Thanks though.

PR #1410, opened 2023-12-01
Add support for OracleSQL

radughiorma:
Added support to use the oracle database.


snicoll:
Thanks for the PR but the docker compose makes it so that the database is provided so there's no need to have two of them.

snicoll:
Sorry, I hadn't noticed we already do this for mysql/postgresql.

@dsyer is that really expected?

dsyer:
We always (before docker) had Postgres and MySql, so we just continued the tradition I think. I’m not sure I’m happy having every single database in the universe (just because the jar files mount up and most people don’t use more than one). For my part I’m am OK with closing this.

PR #1428, opened 2023-12-29
Refactor: Use Environment Variables for URLs and @Autowired for Dependency Injection 

gabemagioli:
Replaced hardcoded URLs "owners/createOrUpdateOwnerForm" and "pets/createOrUpdatePetForm" with environment variables, enhancing flexibility.

Removed the static final type for variables to enable injection, allowing for easier configuration changes.

Replaced constructor injection for OwnerRepository with the @Autowired annotation, reducing lines of code and improving readability.


snicoll:
Thanks for the PR.

> Replaced hardcoded URLs "owners/createOrUpdateOwnerForm" and "pets/createOrUpdatePetForm" with environment variables, enhancing flexibility.

There's actually no requirement for being able to customize views like this.

> Replaced constructor injection for OwnerRepository with the @Autowired annotation, reducing lines of code and improving readability.

Actually, that isn't a best practice. First of all, the field is not `final` as it should be and it makes testing of the component harder.

PR #1361, opened 2023-09-26
Added user friendly messages if blank fields

Aayush-Kapur:
BEFORE :
image

AFTER : 
Screenshot 2023-09-26 at 3 03 16 PM

Screenshot 2023-09-26 at 3 03 39 PM

Screenshot 2023-09-26 at 3 03 59 PM

image

Issue #1351, opened 2023-09-13
Telephone number validation is not user friendly while adding owner


Jitendrajadhav1990:
while adding owner, getting validation for Telephone is "  numeric value out of bounds (<10 digits>.<0 digits> expected)
must not be blank" which will can be user friendly like "Please enter a valid 10 digit mobile number
![Screenshot (1370)](https://github.com/spring-projects/spring-petclinic/assets/144873384/d3ad2f7f-d2b7-4d07-a0c6-318900070b99)

dsyer:
That one should be trivial to fix. It's just a message isn't it? You could start with a PR for the default (English).

Aayush-Kapur:
Hey @dsyer , I've Added User Friendly Messages for blank fields. Please review this PR #1361 .
I've also attached screenshots of before and after for better understanding of output of my commit.

codespearhead:
@Jitendrajadhav1990 It seems that #1361 (merged by accident, now reverted) fixes the same issue as #1483, which was merged three week ago. Should this issue be closed?

PR #1442, opened 2024-02-05
Add user friendly validation messages

lukasdo:
As described in #1351 the validation messages are not very user-friendly. Additionally the validation of the telephone number (which is a String value) with `@Digits` is faulty because leading zeros are ignored when converting from String to BigDecimal in the [DigitsValidator](https://github.com/hibernate/hibernate-validator/blob/7.0.5.Final/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/DigitsValidatorForCharSequence.java#L60). This faulty behaviour was also visible in the [tests](https://github.com/spring-projects/spring-petclinic/blob/836d111e9950f6abedb4fef7b4a94941b8dfedd8/src/test/java/org/springframework/samples/petclinic/owner/OwnerControllerTests.java#L120) where an 11 digit number was parsed as the telephone number.
```
	@Test
	void testProcessCreationFormSuccess() throws Exception {
		mockMvc
			.perform(post("/owners/new").param("firstName", "Joe")
				.param("lastName", "Bloggs")
				.param("address", "123 Caramel Street")
				.param("city", "London")
				.param("telephone", "01316761638"))
			.andExpect(status().is3xxRedirection());
	}
```
These caused the failing tests in #1361.

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/model/Person.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/Person.java
@@ -28,11 +28,11 @@ import jakarta.validation.constraints.NotBlank;
 public class Person extends BaseEntity {
 
 	@Column(name = "first_name")
-	@NotBlank
+	@NotBlank(message = "First Name cannot be blank")
 	private String firstName;
 
 	@Column(name = "last_name")
-	@NotBlank
+	@NotBlank(message = "Last Name cannot be blank")
 	private String lastName;
 
 	public String getFirstName() {
--- a/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
@@ -18,6 +18,7 @@ package org.springframework.samples.petclinic.owner;
 import java.util.ArrayList;
 import java.util.List;
 
+import jakarta.validation.constraints.Pattern;
 import org.springframework.core.style.ToStringCreator;
 import org.springframework.samples.petclinic.model.Person;
 import org.springframework.util.Assert;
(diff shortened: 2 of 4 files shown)


src/main/java/org/springframework/samples/petclinic/model/Person.java
```diff
@@ -28,11 +28,11 @@
 public class Person extends BaseEntity {
 
 	@Column(name = "first_name")
-	@NotBlank
+	@NotBlank(message = "First Name cannot be blank")
```
zyberzebra, line 31: THe message is redundant and should be removed

src/main/java/org/springframework/samples/petclinic/owner/Owner.java
```diff
@@ -47,16 +48,16 @@
...
 	@Column(name = "city")
-	@NotBlank
+	@NotBlank(message = "City cannot be blank")
 	private String city;
 
 	@Column(name = "telephone")
-	@NotBlank
-	@Digits(fraction = 0, integer = 10)
+	@NotBlank(message = "Telephone cannot be blank")
+	@Pattern(regexp = "\\d{10}", message = "Telephone must be a 10-digit number")
```
zyberzebra, line 60: Great solution !

zyberzebra (changes requested):
Good Catch with the digits! I woud remove the messages you added, because they dont provide any benefit and just add maintenance

src/main/java/org/springframework/samples/petclinic/model/Person.java
lukasdo: The message was requested in #1361

dsyer:
If someone can resolve the conflict I'm happy to merge.

dsyer:
I think this is obsolete with #1483

PR #1444, opened 2024-02-09
Add Gradle Checkstyle

ThomasPerkins1123:
The previous attempt to add a checkstyle to gradle failed, but after more thorough testing this should work.
I also made the paths in the suppression file OS agnostic as previously if you ran `./gradlew build` followed by `mvn install` you would get a checkstyle failure on windows as it would not pick up on the path for the build suppressions.


dsyer:
You had to duplicate the suppressions XML file? Seems like a dangerous move (not that it changes very often, but it will definitely break next time someone updates one of them and didn't know about the other).

PR #1468, opened 2024-03-10
Branch2

ViiKeNDa:


andrzejsydor (changes requested):
Why do you want to delete the licence?

PR #1483, opened 2024-03-20
Add user friendly error message

pedrochevallier:


pedrochevallier:
I just realized it's the same as issue #1442, I removed the leading zeros in the tests and added the message for when the number doesn't have 10 digits.

PR #1498, opened 2024-03-28
Remove unnecessary semicolon in OwnerControllerTests

ykmxxi:
- remove unnecessary semicolon in OwnerControllerTests


dsyer:
I think someone already made this change.

PR #1507, opened 2024-04-08
Fix compile error by adding name attributes to @PathVariable and @RequestParam

kooky1001:
Fix compile error by adding name attributes to @PathVariable and @RequestParam


dsyer:
This should not be necessary. Please check your environment for JDK java compiler arguments not provided by the build scripts here (e.g. in your IDE).

snicoll:
Yes compiling with `-parameters` is what you should be doing.

PR #1514, opened 2024-04-15
Support for JetBrains IDEs in Dev Container 

trukhinyuri:
Currently, support for Dev Containers in the project is exclusively oriented towards users of the vscode editor, which does not allow for development in an IDE. 
Specifically, the devcontainer metadata contains proprietary commands that are only compatible with VSCode. 
The devcontainer.json also uses bind, which is not compatible with using the project on remote Docker servers (in this case, docker volumes should be used, as one cannot assume that the host is always exclusively owned by the user). 

To address this, subfolders for jetbrains and vscode were created (fully in accordance with the containers.dev specification). In the jetbrains folder, all specified scenarios have been corrected. 
The vscode folder retains the original configuration. 

When building a devcontainer, both IntelliJ IDEA 2024.1 and vscode can, in the case of multiple devcontainer.json files in the project, ask the user which metadata file to use, thus editor users will choose VSCode, while IDE users will opt for IntelliJ IDEA.
image


dsyer:
It would be good to get someone from Microsoft to look at this (@jdubois or @roryp maybe?) to make sure that the vscode files still work, and that the experience isn't degraded in Codespaces (which was the main target originally).

(I also think you should be careful not to throw shade on vscode in your comments, but it's possible that's just a non-native English speaker making an honest mistake, so I might have drawn the wrong conclusion.)

jdubois:
Thanks for pinging me @dsyer ! Yes the main idea was to have it work with Codespaces, as everyone has 60 hours free per month, it helps a lot for small contributions to OSS projects, so this needs to continue working.
Anyway, the current configuration needs to be updated, I have much better devcontainer configurations now :-)
I'll have a look with my team at Microsoft!

dsyer:
Ping @jdubois again. Any input most welcome.

jdubois:
I just tried it in Codespaces and it didn't work. I didn't have a popup asking me which Devcontainer to use, and as a result it started with the default Java configuration (using JDK 11).

dsyer:
Can you rebase and force push to include the changes from #1623?

dsyer:
Timeout on rebase. If someone wants to do it just send another PR.

PR #1638, opened 2024-08-13
Utilize active profiles with the help of spring.docker.compose.profiles.active property

kgromov:
In order to avoid manual docker-compose run `spring-boot-docker-compose` scope was extended and `spring.docker.compose.profiles.active` set to accept active profile(s).

So after changes it's possible to run petclinic app just with:
```
 ./mvnw spring-boot:run "-Dspring-boot.run.arguments=--spring.profiles.active=postgres"
```
or with IDE adding corresponsding profile. and let `spring-docker-compose` automatically start up.

@pivotal-cla This is an Obvious Fix

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,7 +121,6 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-docker-compose</artifactId>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.testcontainers</groupId>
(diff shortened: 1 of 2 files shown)


dsyer:
You can get the same behaviour with no changes if you run it as `./mvnw spring-boot:test-run -Dspring-boot.run.main-class=org.springframework.samples.petclinic.PostgresIntegrationTests`. I'm happy to add more notes to the README if you think that would help, but I'm not keen on putting docker-compose in the production runtime classpath.

PR #1651, opened 2024-09-06
Docker compose deployment

pankajpc15:
Implement Docker Deployment with Docker Compose

Description:

This Pull Request introduces Docker deployment for our application using Docker Compose. The following changes have been made:

Added Docker Compose Configuration:

Created a docker-compose.yml file to define services, networks, and volumes for the application.
Configured services for the application, including any necessary environment variables and dependencies.
Updated Dockerfile:

Adjusted the Dockerfile to be compatible with the Docker Compose setup.
Tested Deployment:

Verified that the Docker Compose setup works correctly by running and testing the containers locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM openjdk:17-jdk-slim
+
+WORKDIR /app
+
+# Copy the application JAR file 
+COPY spring-petclinic-3.3.0-SNAPSHOT.jar /app/spring-petclinic.jar
+
+EXPOSE 8080
+
+ENTRYPOINT ["java", "-jar", "spring-petclinic.jar"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3.8'
 services:
   app:
     image: spring-petclinic:latest
-    container_name: app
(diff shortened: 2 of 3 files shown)


pankajpc15:
@pivotal-cla sign agreement successfully

pankajpc15:
@pivotal-cla  can you accept PR

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM openjdk:17-jdk-slim
```
andrzejsydor, line 1: Why not a never one like 21/22 jdk version?

pankajpc15: @andrzejsydor 
Great observation! I chose JDK 17 because it's a long-term support (LTS) release, ensuring stability and extended support. Additionally, the project code is fully compatible with Java 17, so I opted for it.

dsyer:
I don't think we can merge this because it removes the existing configuration with MySQL. I'm not sure why you would prefer to launch the app in a container with docker-compose when it is running locally, so probably this change is targeted at some platform? Anyway, the local integration tests and dev time application launch (from CLI or IDE) would break if we make these changes.

PR #1654, opened 2024-09-10
Support for Spring AI based chat bot in Petclinic

odedia:
This pull request adds a new Chat Box in Spring Petclinic.
It is used to ask the clinic questions about owners and vets.
Some of the questions you may ask:
1. Please list the owners that come to the clinic.
2. How many vets are there?
3. Is there an owner named betty?
4. Which owners have dogs?
And many more.
You may also try to add a pet to an existing owner - at the moment it seems like that actual pet is not added to the owner, however, the chatbot will guide you through the needed information required in order to create a new pet. This will be enhanced in the future.
The code currently only works under the spring profile `postgres-openai`. It requires Postgres and the azure-openai spring AI dependency.
You will need to copy `creds-template.yaml` to `creds.yaml` and provide your endpoint and API key.


odedia:
Will provide additional enhancements before submitting.

PR #1656, opened 2024-09-11
Support a Spring AI based chatbot in Petclinic

odedia:
This pull request adds a new Chatbot in Spring Petclinic.
It can be used used to ask the clinic questions about owners and vets.
Some of the questions you may ask:

1. Please list the owners that come to the clinic.
2. How many vets are there?
3. Is there an owner named Betty?
4. Which owners have dogs?
5. Add a dog for Betty. Its name is Moopsie.

And many more.

The implementation is under  the spring profile `openai`, which is the only supported LLM at the moment (either azure-openai or regular openai). In the default profile, the chatbot returns "chat is unavailable".
See guide in main `readme.md` on how to use.

### DIFF
--- a/src/main/scss/chatbot.scss
+++ b/src/main/scss/chatbot.scss
@@ -1,0 +1,88 @@
+.chatbox {
+  position: fixed;
+  bottom: 0;
+  right: 0;
+  width: 300px;
+  border-radius: 10px 10px 0 0;
+  box-shadow: 0 4px 8px rgba(0, 0, 0, 0.2);
+  font-family: Arial, sans-serif;
+  background-color: #f0f0f0;
+  overflow: hidden;
+  
+  &.minimized {
+    height: 40px;
+  }
+
+  &-header {
+    background-color: $spring-green;
+    color: white;
+    padding: 10px;
+    cursor: pointer;
+    font-weight: bold;
+    text-align: center;
+  }
+
+  &-content {
+    height: 400px;
+    display: flex;
+    flex-direction: column;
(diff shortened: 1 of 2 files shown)


dsyer:
This looks like really good work and sticks well (but maybe not quite well enough) to the constraints we impose on this project.  The main constraint it violates is "no new features", so I'm not sure we want to merge really, but it's good to see how to do it. There's always the possibility of creating a new project in https://github.com/spring-petclinic.

I'm quite pleased that the Javascript is nearly free of dependencies, but it would be better to use webjars (same as all the others). The CSS changes probably need to go in `petclinic.scss` (there's a reason that `petclinic.css` is ignored by git). You would also need to remove the kubernetes/tanzu specific configuration files, if you feel like pushing it along a bit further.

odedia:
Perhaps it makes more sense to include this in a separate branch and not the main branch?
I’ll address the SCSS issue remove the Tanzu specific code.

dsyer:
A separate branch might be good. It would be unlikely to get any maintenance, but it's probably a good way to evolve the feature (which feels like it might need to mature a bit - e.g. making something that works without any additional configuration maybe, or without a subscription at least).

dsyer:
I created a new branch `spring-ai`. Please submit more PRs against that (e.g. some documentation about how to set up the credentials, maybe a chat session example).

odedia:
Thanks, will do. I added documentation in the main `readme.md` but it can always be improved.

dsyer:
Nice! I missed that somehow.

Issue #1664, opened 2024-09-24
Small Code Cleanup


mosbat:
I inspected code and found the following:
1- Unused imports.
2- Made some variables final since they are part of DTOs.

dsyer:
No CLA in PR

mosbat:
> No CLA in PR

Thank you for the feedback. I think that I signed the CLA for the PR but please don't mind me if I missed something. I'm learning.  Btw, I see the CLA signed in the PR's checks...

dsyer:
Sorry, I think I was looking at the wrong PR. It’s merged now.

PR #1665, opened 2024-09-24
Minor Code Cleanup

mosbat:
#1664 

Did some minor code cleanup


mosbat:
After the changes I made, I ran all the existing tests and was able to successfully build and run the application locally.

andrzejsydor:
The PR is not just about remove unused imports.
Should be updated the commit message, imo

mosbat:
@andrzejsydor thank you for the feedback; if you have a look at again, there is a line break that also states the 2nd change. However, if you couldn't see it, it means that I failed to make the commit message clearer :)

Issue #1671, opened 2024-09-26
New Feature - View Vet Profile 


mosbat:
A new feature I propose, to further extend the features in future in relation to Vets' management, the admin can view the Vet's profile and potentially change some other details in future (in another PR or issue).

If this will be approved, then the next step will be adding more details related to the vets and possibility to edit some of their details.

Thank you for reviewing my changes and as always, I'm open for ideas, suggestions or improvements!

PR:
https://github.com/spring-projects/spring-petclinic/pull/1672

mosbat:
Screenshots of UI changes

![image](https://github.com/user-attachments/assets/a4edeafe-4cb2-4f2d-af4a-f23e27bff2f5)


![image](https://github.com/user-attachments/assets/30635c16-a965-4f61-bc70-81f69bd4afc6)

PR #1672, opened 2024-09-26
New Feature - #1671 - Added Profile Page for Vets

mosbat:
#1671


mosbat:
> The effort is appreciated, but I'm not sure we need this (and the usual policy on new features is to decline). Also the UUID change is kind of unnecessary as well as being invasive.

No worries, I didn't know about the policy :) but thanks for letting me know! I had some time to kill :).

You're right that the UUID change is not necessary; probably overkill. 

If you like I can close this PR and work a bit on the existing code base with smaller changes :) .

dsyer:
Contributions always gratefully accepted.

PR #1683, opened 2024-10-12
added Russian language support

ogbozoyan:


src/main/resources/messages/messages_ru.properties
```diff
@@ -0,0 +1,8 @@
+welcome=Добро пожаловвать
```
dsyer, line 1: Google translate gives me a different spelling. Do we need 2 "в"?

ogbozoyan: Thanks, it's just a typo. I fixed it.

dsyer:
Could you please squash your commits into one and force push?

PR #1691, opened 2024-10-19
feat: Adds support for the Portuguese language.

joaobertholino:
Adding support for the Portuguese language.

### DIFF
--- a/src/main/resources/messages/messages_pt.properties
+++ b/src/main/resources/messages/messages_pt.properties
@@ -1,0 +1,8 @@
+welcome=Bem-vindo
+required=É necessário
+notFound=Não foi encontrado
+duplicate=Já está em uso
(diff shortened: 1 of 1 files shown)


src/main/resources/messages/messages_pt.properties
```diff
@@ -0,0 +1,8 @@
+welcome=Bem-vindo
+required=� necess�rio
```
dsyer, line 2: I think the binary bytes are wrong here? They should show up as accented letters and I'm just seeing (?).

joaobertholino: I could remove the accentuation, but it wouldn't be in accordance with the correct Portuguese language.
Could you attach a printout of how it appears to you, please?

Issue #968, opened 2022-05-15
Add deployment for spring boot in k8s


shivampateriya1:
I want to contribute in the spring-pet clinic by adding k8s deployment file and SQL file in k8s.

dsyer:
We've talked about it before but never really had a consensus on what should go into the manifest. There are lots of choices and not many are obviously right.

xerx593:
> I want to contribute in the spring-pet clinic by adding k8s deployment file and SQL file in k8s.

I just want to mention: https://github.com/spring-petclinic/spring-petclinic-cloud (official fork)

dsyer:
That would be a useful reference for anyone wanting to contribute. Also see here for some working example code: https://github.com/dsyer/spring-petclinic/tree/k8s.

I would say we need as few lines of YAML as we can get away with using just vanilla `kubectl` and Spring Boot features.

0) Use the image generated by `mvn spring-boot:build-image` and deploy either in a local repository or a public one (like dockerhub)
1) minimal service and deployment
2) just one database (probably mysql) and a config map following the [Service Binding](https://github.com/servicebinding/spec) spec because Spring Boot will pick that up automatically
3) maybe `/livez` and `/readyz` (can be exposed through a system property or application property)

panic08:
> That would be a useful reference for anyone wanting to contribute. Also see here for some working example code: https://github.com/dsyer/spring-petclinic/tree/k8s.
> 
> I would say we need as few lines of YAML as we can get away with using just vanilla `kubectl` and Spring Boot features.
> 
> 0. Use the image generated by `mvn spring-boot:build-image` and deploy either in a local repository or a public one (like dockerhub)
> 1. minimal service and deployment
> 2. just one database (probably mysql) and a config map following the [Service Binding](https://github.com/servicebinding/spec) spec because Spring Boot will pick that up automatically
> 3. maybe `/livez` and `/readyz` (can be exposed through a system property or application property)

Didn't you implement this functionality in [your](https://github.com/dsyer/spring-petclinic/tree/k8s) fork?

dsyer:
Yes I did. But I don’t maintain it, and no-one ever gave any feedback.

dsyer:
Merged #1707

PR #1707, opened 2024-11-07
Add Kubernetes support

panic08:
I made some changes that are aimed at adding Kubernetes support to spring-petclinic, for a quick deploy using Kubernetes

I added manifests like `db.yml` and `petclinic.yml`

The first manifest contains the secrets for the database, and the service and deployment. I used Postgresql 17 as the database

The second manifest contains the service and deployment with 1 replica for directly our petclinic. I configured service-bindings, also “liveness” endpoints like `/livez` and `/readyz` were added. The [dsyer/petclinic](https://hub.docker.com/r/dsyer/petclinic) was used as the image

This PR is related to issue #968. Also, I was guided by this [branch](https://github.com/dsyer/spring-petclinic/tree/k8s)

### DIFF
--- a/k8s/db.yml
+++ b/k8s/db.yml
@@ -1,0 +1,73 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: demo-db
+type: servicebinding.io/postgresql
+stringData:
+  type: "postgresql"
+  provider: "postgresql"
+  host: "demo-db"
+  port: "5432"
+  database: "petclinic"
+  username: "user"
+  password: "pass"
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: demo-db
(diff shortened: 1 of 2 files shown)


andrzejsydor (approved):

dsyer:
This looks good, thank you, and thank you to @andrzejsydor for approving. Other opinions would be valuable if you have anything to say, even if it's just a thumbs up.

I'm not super excited about being the host of the docker image, but it's a good first step. I'll see if we can get a github action to push to spring-io/spring-petclinic (or something), but that can happen later.

Another thing that's missing is a test, but that's kind of a big ask and again not mandatory for a first draft. If anyone wants to play with github actions using Kind (my preference) that would be a welcome contribution.

PR #1710, opened 2024-11-08
Implement K8S deployment testing using Kind in GitHub Actions

panic08:
Essentially this PR is a reference to @dsyer [comment](https://github.com/spring-projects/spring-petclinic/pull/1707#issuecomment-2464215589), more specifically the advice to create a separate workflow for K8S deployment

Here I have created a special workflow that deploys the database and our spring-petclinic, and checks if it has successfully deployed or not

### DIFF
--- a/.github/workflows/deploy-and-test-cluster.yml
+++ b/.github/workflows/deploy-and-test-cluster.yml
@@ -1,0 +1,31 @@
+name: Deploy and Test Cluster
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'k8s/**'
+  pull_request:
+    branches: [main]
+    paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-and-test-cluster.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy and Test Cluster
+
+on:
+  push:
+    branches: [main]
```
dsyer, line 5: Maybe this could be filtered by path? `paths: - k8s/**` for example.

panic08: Yeah, that's right, thanks

PR #1717, opened 2024-11-17
 Add "Open with Codeanywhere" badge

ivan-burazin:
### Add "Open with Codeanywhere" Badge

Why? With Codeanywhere, contributors can:

1. **Automatically set up and configure a remote dev environments** if a `devcontainer.json` file is present (such as this repo).
2. **Instantly launch a cloud or local IDE** to the setup environment.

### Benefits:
- Faster onboarding for new contributors.
- Consistent and ready-to-use development environments.
- Flexibility to work in the cloud or locally.


dsyer:
I like it. I'm not sure if we should have this link though (also the Gitpod one really) since it requires unlimited repository access in Github, and that's not something I want people to give away lightly, or with an implied recommendation from the PetClinic team.

ivan-burazin:
Hi @dsyer,
Thanks for getting back! The only thing we can do using the OAuth API (https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps) is potentially offer the possibility to enable access only to public repositories.

Would that be an adequate solution?

dsyer:
I think the problem is requiring write access to any repo (public or private). I don't want to give random third parties those kinds of permissions, and I don't want to encourage other people to do that either. I don't need write access to anything (except some storage) to explore a clone of a public repo in my own environment. I appreciate you don't have many options, and GitHub is making things hard for you.

ivan-burazin:
How would you edit it in a Cloud environment without write access? I understand the worries about private repo, but what would be the danger's of public?

I completely understand that you can say no to this PR, but honestly interested.

dsyer:
> How would you edit it in a Cloud environment without write access?

You can edit a clone without any special access, and that might be enough to get started. You can't commit it without a private key, or an access token like you are asking for. That's where things get sticky I guess. The access is too high to grant to an unknown 3rd party.

PR #1740, opened 2024-12-14
 space added for build gradle

Chimakurthi7788:
space added for build gradle


Chimakurthi7788 (commented):
need to correct code

andrzejsydor (changes requested):
no changes at all

PR #1753, opened 2024-12-20
Pet update fix

smnk107:
I have made a little change in side processUpdateForm api, in petController file. To update the pet details for an owner, we are iterating the existing pet list and if any pet id matches with the changed pet id, we are replacing that pet details with the updated details.


src/main/java/org/springframework/samples/petclinic/owner/PetController.java
```diff
@@ -146,10 +147,34 @@ public String processUpdateForm(Owner owner, @Valid Pet pet, BindingResult resul
 			return VIEWS_PETS_CREATE_OR_UPDATE_FORM;
 		}
 
+<<<<<<< HEAD
+		owner.addPet(pet);
+		owner.updatePet(pet);
+
+||||||| 6148ddd
```
andrzejsydor, line 154: It's not ready to merge

PR #1756, opened 2024-12-22
added updatePet in Owner Class

smnk107:
added updatePet method in Owner Class to update the pet and have called it from PetController.

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
@@ -172,4 +172,17 @@ public class Owner extends Person {
 		pet.addVisit(visit);
 	}
 
+	public void updatePet(Pet newPet)
+	{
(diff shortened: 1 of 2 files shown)


abdurraffaysyed (approved):

src/main/java/org/springframework/samples/petclinic/owner/Owner.java
```diff
@@ -172,4 +172,16 @@ public void addVisit(Integer petId, Visit visit) {
 		pet.addVisit(visit);
 	}
 
+	public void updatePet(Pet newPet)
+	{
+		Pet existingPet = getPet(newPet.getId());
```
andrzejsydor, line 177: imo, if there is no pet then some exception. The client want to update the pet, but what if no exists, then no info of that.

dsyer:
Is this actually necessary? A test that proves it might be a good idea. I assume it's probably just a doodle, but please scream if you think otherwise.

PR #1758, opened 2024-12-25
edited version (AhmedMagdyy)

AhmedMagdyy-eng:


dsyer:
Some good bits. Plenty of unnecessary or unuseful bits. Break it up if you feel deeply about any of it, please.

PR #1762, opened 2024-12-27
Refactor changes and code smells

abdelrahmanassem:
Refactor changes and code smells  listed : 

1. long method
2. primitive obsession
3. code duplication
4. unnecessary comments

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
@@ -103,10 +103,7 @@ class PetController {
 		if (StringUtils.hasText(pet.getName()) && pet.isNew() && owner.getPet(pet.getName(), true) != null)
 			result.rejectValue("name", "duplicate", "already exists");
 
-		LocalDate currentDate = LocalDate.now();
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/samples/petclinic/owner/PetController.java
kalyaniarla, line 1: owner.getPet(petId) in line 79 need to be updated to use the findPet method instead

src/main/java/org/springframework/samples/petclinic/owner/PetIdentifier.java
```diff
@@ -0,0 +1,21 @@
...
+	public PetIdentifier(String name) {
+		this.name = name;
+		this.id = null;
+	}
+
+	public boolean matches(Pet pet) {
+		if (id != null) {
+			return pet.getId().equals(id);
+		}
+		return pet.getName().equalsIgnoreCase(name);
```
andrzejsydor, line 19: That's mean the pet name is unique, but is it ok?

dsyer:
I see what you want to do. I'm not sure it's justified. Thanks.

Issue #1752, opened 2024-12-19
Pet details are not updated


99MMRR99:
Example: go to first owner with pet Leo of type cat.
Click on edit pet and change the type to bird or change the name.
Click on Update pet.
The pet is not updated, even though a success message is shown.
When I analysed the code, I found that the updated pet is given to the addPet method of the owner which does not change the pet in the owner list, as the pet is not new.
In PetController:
![grafik](https://github.com/user-attachments/assets/3d33ef4c-5340-4dfc-8e1d-cdb1847d0df6)

smnk107:
Hey !
I would like to work on this issue. Could you please assign it to me?

99MMRR99:
It seems that I cannot assign anybody to the issue, I dont know why, but feel free to fix the issue.

smnk107:
Okey !

smnk107:
Hello !
I have submitted a PR for this issue. Please review it when you have the time. Let me know if further improvements are needed. Thanks!

99MMRR99:
The solution is good and solves the problem. I would only suggest to put the whole logic with the for loop into a new method in the Owner (maybe call it updatePet). And I am not sure why you renamed the PetController to privatePetController.
Otherwise good work, thank you for fixing it!

smnk107:
Okey...sure ! I will correct the things.

smnk107:
I have created a new PR. Please review it !

abdurraffaysyed:
Thank you for your contribution! The PR looks great and is in good shape overall. I have one suggestion: it's a good practice to avoid leaving commented-out code in the codebase. Removing the commented code in the "PetController" class would help maintain a cleaner code state. Let me know if you have any questions about this!

smnk107:
Thanks for checking this out. I have removed the commented code. Please review it when you have the time. Let me know if further improvements are needed. Thanks!

abdurraffaysyed:
Hi @smnk107,

Thank you for the recent changes! The PR looks great overall, and I appreciate your work on the updatePet() implementation in the Owner class. Upon reviewing the code again, I noticed that the function iterates over the pets list and compares each Pet object with the updated pet object using the pet_Id.

To make the implementation cleaner and more efficient, I suggest leveraging the existing getPet(Integer id) method within the updatePet() function. By doing so, you can directly retrieve the relevant Pet object and replace it with the updated version, avoiding the need for a loop.

[...]

PR #1765, opened 2024-12-31
Fix pet update functionality - Fixed issue where editing a pet's type…

VishantOwO:
# Fix Pet Update Functionality

## Description
Fixed an issue where editing a pet's type or name wasn't persisting in the database, even though a success message was shown.

## Changes made
- Modified `processUpdateForm` to update existing pet properties instead of using `addPet`
- Added proper handling of existing pet updates
- Maintained all existing validation checks

Note: I'm new to contributing and haven't added tests yet. I'd be happy to add them if required.

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
@@ -146,23 +146,26 @@ class PetController {
 			return VIEWS_PETS_CREATE_OR_UPDATE_FORM;
 		}
 
+		updatePetDetails(owner, pet);
+		redirectAttributes.addFlashAttribute("message", "Pet details has been edited");
+		return "redirect:/owners/{ownerId}";
+	}
 
+	/**
+	 * Updates the pet details if it exists or adds a new pet to the owner.
+	 * @param owner The owner of the pet
+	 * @param pet The pet with updated details
+	 */
+	private void updatePetDetails(Owner owner, Pet pet) {
 		Pet existingPet = owner.getPet(pet.getId());
 		if (existingPet != null) {
 			// Update existing pet's properties
 			existingPet.setName(pet.getName());
 			existingPet.setBirthDate(pet.getBirthDate());
 			existingPet.setType(pet.getType());
-
-			this.owners.save(owner);
 		} else {
-
 			owner.addPet(pet);
-			this.owners.save(owner);
 		}
-
-		redirectAttributes.addFlashAttribute("message", "Pet details has been edited");
-		return "redirect:/owners/{ownerId}";
+		this.owners.save(owner);
 	}
-
 }
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
@@ -125,11 +125,11 @@ class PetController {
 
 	@PostMapping("/pets/{petId}/edit")
 	public String processUpdateForm(Owner owner, @Valid Pet pet, BindingResult result,
-			RedirectAttributes redirectAttributes) {
+									RedirectAttributes redirectAttributes) {
 
 		String petName = pet.getName();
 
-		// checking if the pet name already exist for the owner
+		// checking if the pet name already exists for the owner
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/samples/petclinic/owner/PetController.java
```diff
@@ -146,8 +146,21 @@ public String processUpdateForm(Owner owner, @Valid Pet pet, BindingResult resul
 			return VIEWS_PETS_CREATE_OR_UPDATE_FORM;
 		}
 
-		owner.addPet(pet);
-		this.owners.save(owner);
+
+		Pet existingPet = owner.getPet(pet.getId());
+		if (existingPet != null) {
```
abdurraffaysyed, line 151: @VishantOwO 
Could you please extract out the details inside if block to an update function. It will improve the code structure and it's readability.

src/main/java/org/springframework/samples/petclinic/owner/PetController.java
```diff
@@ -146,8 +146,21 @@ public String processUpdateForm(Owner owner, @Valid Pet pet, BindingResult resul
...
+			// Update existing pet's properties
+			existingPet.setName(pet.getName());
+			existingPet.setBirthDate(pet.getBirthDate());
+			existingPet.setType(pet.getType());
+
+			this.owners.save(owner);
+		} else {
+
```
andrzejsydor, line 159: the empty line is not needed

src/main/java/org/springframework/samples/petclinic/owner/PetController.java
VishantOwO: I've extracted the update logic into a separate method updatePetDetails as suggested. The code should be cleaner and more readable now. Please let me know if you'd like any further adjustments.

dsyer:
THank you for the contribution. I believe this is in relation to #1752. Instead of adding a new method wouldn't it be better to fix the `addPet()` method so it updates an existing pet? Also, please add a test.

arey:
I've got the same issue.
From my testing, this problem comes from disabling JPA Open Session In View https://github.com/spring-projects/spring-petclinic/pull/1667

abdurraffaysyed:
@dsyer,

I recently read your comment regarding modifying the addPet functionality to accommodate updates as well. I truly appreciate your insights on this enhancement.

However, I had a small concern regarding this approach. Wouldn’t combining both add and update functionalities in a single method violate the Single Responsibility Principle (SRP)? Ideally, addPet should focus solely on adding a pet, while updates should be handled by a separate function to maintain clear separation of concerns.

I might be overlooking something here, so I would love to hear your thoughts on this. Looking forward to your insights!

arey:
Hi @abdurraffaysyed and @dsyer 
In my opinion, the `updatePetDetails` code should be moved to the `Owner` class. We have the choice of declaring two methods (`add` and `update`) or changing the `addPet` method to a more generic business name. If we keep a single method, I suggest `registerPet`.  The name`upsertPet` looks like better but too technical.

```java
	public void registerPet(Pet pet) {
		Pet existingPet = getPet(pet.getId());
		if (existingPet == null) {
			getPets().add(pet);
		} else {
			existingPet.setName(pet.getName());
			existingPet.setBirthDate(pet.getBirthDate());
			existingPet.setType(pet.getType());
		}
	}

```

src/main/java/org/springframework/samples/petclinic/owner/PetController.java
```diff
@@ -146,10 +146,26 @@ public String processUpdateForm(Owner owner, @Valid Pet pet, BindingResult resul
...
 
+	/**
+	 * Updates the pet details if it exists or adds a new pet to the owner.
+	 * @param owner The owner of the pet
+	 * @param pet The pet with updated details
+	 */
+	private void updatePetDetails(Owner owner, Pet pet) {
+		Pet existingPet = owner.getPet(pet.getId());
+		if (existingPet != null) {
+			// Update existing pet's properties
```
andrzejsydor, line 162: The comment is not needed, imo

andrzejsydor (approved):

dsyer:
Merged in 9f7aa2c thanks.

PR #1767, opened 2025-01-04
ORGANIZED CLASSES INTO PACKAGES

sapphire0602:
Arranged similar classes into packages in the owner package , and also made the controller classes 'public' , for accessibility in the test classes.


dsyer:
I don't see why this is necessary. Also uses unconventional package name.

PR #1783, opened 2025-01-28
Update README.md

sharvari-devops:


andrzejsydor (changes requested):
not needed such a change, imo

PR #1788, opened 2025-02-10
Fix: Add improvements in PetClinic

sukhminderarora:
Introduced Error messages during validations to give more clarity. 
Introduced constants
Modified the inputField.html to remove the non required tags

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/model/Person.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/Person.java
@@ -28,11 +28,11 @@ import jakarta.validation.constraints.NotBlank;
 public class Person extends BaseEntity {
 
 	@Column(name = "first_name")
-	@NotBlank
+	@NotBlank(message = "First name is required")
 	private String firstName;
 
 	@Column(name = "last_name")
-	@NotBlank
+	@NotBlank(message = "Last name is required")
 	private String lastName;
 
 	public String getFirstName() {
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
```diff
@@ -47,6 +47,7 @@
 class OwnerController {
 
 	private static final String VIEWS_OWNER_CREATE_OR_UPDATE_FORM = "owners/createOrUpdateOwnerForm";
+	public static final String REDIRECT_OWNERS = "redirect:/owners/";
```
andrzejsydor, line 50:
```suggestion
	private static final String REDIRECT_OWNERS = "redirect:/owners/";
```

src/main/java/org/springframework/samples/petclinic/model/Person.java
```diff
@@ -28,11 +28,11 @@
 public class Person extends BaseEntity {
 
 	@Column(name = "first_name")
-	@NotBlank
+	@NotBlank(message = "First name is required")
 	private String firstName;
 
 	@Column(name = "last_name")
-	@NotBlank
+	@NotBlank(message = "Last name is required")
```
andrzejsydor, line 35: Better to use translated files, what do you think?

andrzejsydor (changes requested):

src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
sukhminderarora: Thank you for you feedback. You are right. I have made the change accordingly

src/main/java/org/springframework/samples/petclinic/model/Person.java
sukhminderarora: I have included this in the message-properties

dsyer:
I merged but had to revert because of CI failures. Please run the build before you push a commit to a PR. If you want to try again open a new PR, sorry.

PR #1804, opened 2025-02-28
Branch1

Ykianik089:


README.md
```diff
@@ -161,3 +161,12 @@ For additional details, please refer to the blog post [Hello DCO, Goodbye CLA: S
 ## License
 
 The Spring PetClinic sample application is released under version 2.0 of the [Apache License](https://www.apache.org/licenses/LICENSE-2.0).
+Change 1 br1
```
andrzejsydor, line 164: What for is that?

andrzejsydor (changes requested):
The change is not needed, imo

PR #1819, opened 2025-03-13
Removal of the showResourcesVetList method dead code

arey:
Cause no more JSON and XML export feature


dsyer:
Why is it dead code? Is there a bug in the `@GetMapping`?

arey:
The UI menu references the GET `/vets.html?page=1` endpoint (within the `showVetList` method of the `VetController`), not the `/vets` one. Thus  I think we could remove the unused `showResourcesVetList` method.

dsyer:
The UI is clearly meant to show HTML. The `/vets` endpoint still works (and gives you JSON or XML depending on what youask it for). I think that was the purpose of the code that you proposed to remove, so I think it's worth keeping.

dsyer:
Maybe we should put the links back? Or at least add some instructions about how to get the XML and JSON?

PR #1834, opened 2025-03-28
Hilal readme

hilaltengilim:
adımı yazdm


andrzejsydor (changes requested):
Why such a pr? Is it needed?

PR #1839, opened 2025-04-03
add explicit names to @RequestParam and @PathVariable annotations

Nirsa-Dev:
to prevent runtime exception

Spring Framework 6+ requires parameter name reflection or explicitly named annotations.
This commit adds missing `name = "..."` attributes to all @RequestParam and @PathVariable
usages across controller classes.

Affected controllers
- OwnerController
- PetController
- VisitController
- VetController

This change prevents the following runtime error : IllegalArgumentException: Name for argument of type [int] not specified...

The application now builds and runs cleanly with `./gradlew clean build`.


andrzejsydor (changes requested):
The explicit names are not needed here.

src/test/java/org/springframework/samples/petclinic/SpringPetclinicApplicationTests.java
```diff
@@ -0,0 +1,13 @@
+package org.springframework.samples.petclinic;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+class SpringPetclinicApplicationTests {
```
andrzejsydor, line 7: why empty tests?

src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
```diff
@@ -32,9 +32,9 @@
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.mvc.support.RedirectAttributes;
```
andrzejsydor, line 35: why changed the imports, not needed

src/main/java/org/springframework/samples/petclinic/SpringPetclinicApplication.java
```diff
@@ -0,0 +1,13 @@
+package org.springframework.samples.petclinic;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class SpringPetclinicApplication {
```
andrzejsydor, line 7: Why a new Applicatin class?

andrzejsydor (changes requested):

Nirsa-Dev:
Due to an Eclipse bug, an unintended push occurred and the project was accidentally deleted.
Some experimental code I was playing around with also got pushed by mistake.

From what I can tell, only the page parameters actually require the name attribute in @RequestParam.

What are your thoughts on this?
If you agree with this approach, I’ll resolve the issue on my machine and submit a clean PR accordingly.

Thank you for your understanding

PR #1842, opened 2025-04-07
updated pom.xml

keerthima597:


andrzejsydor (changes requested):
Not needed such a change

PR #1845, opened 2025-04-07
"Updated welcome message for DevOps practice"

Sumerakhan532:


src/main/resources/templates/welcome.html
```diff
@@ -4,7 +4,7 @@
 
   <body>
 
-    <h2 th:text="#{welcome}">Welcome</h2>
+    <h2 th:text="#{welcome}">Welcome to Devops PetClinic Page</h2>
```
andrzejsydor, line 7: Why `devops ....` page?

andrzejsydor (changes requested):

PR #1846, opened 2025-04-07
Message changes in welcome page

Sumerakhan532:


src/main/resources/templates/welcome.html
```diff
@@ -4,7 +4,7 @@
 
   <body>
 
-    <h2 th:text="#{welcome}">Welcome</h2>
+    <h2 th:text="#{welcome}">Welcome to Devops Pet</h2>
```
andrzejsydor, line 7: why `devops` ?

andrzejsydor (changes requested):

PR #1848, opened 2025-04-07
Frogbot2

JesseHouldsworth:


andrzejsydor (changes requested):
What is 'Frogbot'?
Why so many commits, better to squash before merge it .

PR #1852, opened 2025-04-09
Fix/utop bot 90b6fea1

amrohendawi:


dsyer:
I don't think we want any of the pom.xml changes (especially not removing perfectly good comments). The commit log needs fixing (per DCO requirements). The test changes might be interesting, but not high priority, so I'll wait for some changes and then probably close if nothing happens.

amrohendawi:
I am developing an AI agent to fix weak tests in java projects. This PR was mistakenly opened by the agent under development.

PR #1866, opened 2025-04-26
Edit build gradle for building

stella-03:
adding printBuildInfo, Jacoco in build.gradle


src/.DS_Store
andrzejsydor, line 1: The file shouldn't be merged, pleas remove it

andrzejsydor (changes requested):
Please remove macos specific file

PR #1878, opened 2025-05-01
Internationalization Enhancement

anizmo:
## Commit Contents:
- Added test to ensure there are no string literals
- Added test to ensure a string is translated in all language files
- Added missing strings in properties
- Internationalized remaining strings flagged by the tests

## Context:
- This makes it easier to enforce that we always follow internationalization, and a preventive approach works better than searching and fixing individual strings later.
- This test flagged each missing string for me, and I was able to resolve each one of them.


anizmo:
Could someone please assist me in merging this? This one is approved, but I don't see an option to merge it.

PR #1884, opened 2025-05-07
Added endpoints, model, schema, and services for pet attributes

koushaljain:
…ttributes"


dsyer:
Author seems uninterested in fixing CI, and I'm not sure we need another table in the database. The naming convention for packages doesn't match the one in the project either, so on the whole I think this is probably not a serious proposal.

PR #1918, opened 2025-06-11
Add test case for invalid phone number format in owner creation form

shaileshkumargupta:


dsyer:
I have no objection to this. You would need to sign the commit though (per the failing action).

PR #1956, opened 2025-07-16
Enhancement: Added Lombok for model in place of getters and setter

Nishcahy:
## Summary
 
Refactored the codebase to use Lombok annotations for cleaner and more maintainable code.
 
## Changes
 
- Added Lombok dependency in `pom.xml`
- Replaced manual constructors with `@AllArgsConstructor`
- Removed explicit getters and setters, added `@Getter` and `@Setter`
- Applied to relevant classes (e.g., Service, Controller)
 
## Motivation
 
Using Lombok reduces boilerplate and improves readability while supporting constructor-based dependency injection in Spring Boot.
 
## Checklist
 
- [x] Changes tested locally
- [x] Application builds successfully
- [x] No breaking changes

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/model/BaseEntity.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/BaseEntity.java
@@ -21,8 +21,8 @@ import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
 import jakarta.persistence.MappedSuperclass;
-import lombok.Data;
 import lombok.Getter;
+import lombok.Setter;
 
 /**
  * Simple JavaBean domain object with an id property. Used as a base class for objects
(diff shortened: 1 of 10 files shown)


src/main/java/org/springframework/samples/petclinic/model/BaseEntity.java
```diff
@@ -21,6 +21,8 @@
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
 import jakarta.persistence.MappedSuperclass;
+import lombok.Data;
```
andrzejsydor, line 24: The import is not needed

src/main/java/org/springframework/samples/petclinic/model/BaseEntity.java
```diff
@@ -30,19 +32,13 @@
  * @author Juergen Hoeller
  */
 @MappedSuperclass
+@Data
```
andrzejsydor, line 35: What about equals and hashCode? Here you have baseentity

andrzejsydor (changes requested):
still the code can be improved

PR #1969, opened 2025-07-23
Test controller with review issues

mkierebi:


src/main/java/org/springframework/samples/petclinic/system/TestController.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class TestController {
+
+    @GetMapping("/test")
+    public String test(String input) {
+        // This code has potential issues for testing the review system
+        if (input == null) {
+            return null; // Should probably throw exception or return default
```
mkierebi, line 13: Should throw IllegalArgumentException instead of returning null

src/main/java/org/springframework/samples/petclinic/system/TestController.java
```diff
@@ -0,0 +1,25 @@
...
+        }
+        
+        // No input validation
+        return "Hello " + input;
```
mkierebi, line 17: should check for null before calling toLowerCase()

PR #1990, opened 2025-08-11
Add OpenAPI docs and monitoring with Prometheus and Docker Compose

RAJESHREDDY0508:
This pull request introduces OpenAPI/Swagger documentation to the PetClinic application using springdoc-openapi, adds a configuration class, and updates the build accordingly. It also exposes Prometheus metrics via the Spring Boot actuator and adds a docker-compose.dev.yml file to spin up Postgres, Prometheus, Grafana, and the PetClinic service for local development. The README has been updated with quick start instructions and endpoints for Swagger UI, Prometheus, and Grafana.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,12 @@
 
   <build>
     <plugins>
+     <!-- API documentation -->
+        <dependency>
+            <groupId>org.springdoc</groupId>
+            <ar>tifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
+            <version>2.6.0</version>
+        </dependency>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -150,6 +150,12 @@
 
   <build>
     <plugins>
+     <!-- API documentation -->
+        <dependency>
```
andrzejsydor, line 154: it's plugin section, why here is the dependency?

src/main/java/org/springframework/samples/petclinic/config/OpenApiConfig.java
```diff
@@ -0,0 +1,10 @@
+package org.springframework.samples.petclinic.config;
+
+import io.swagger.v3.oas.annotations.OpenAPIDefinition;
+import io.swagger.v3.oas.annotations.info.Info;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@OpenAPIDefinition(info = @Info(title = "Pet Clinic API", version = "1.0", description = "Documentation for Pet Clinic API"))
+public class OpenApiConfig {
```
andrzejsydor, line 9:
```suggestion
class OpenApiConfig {
```

andrzejsydor (changes requested):
wrong Logging level

PR #1993, opened 2025-08-14
Feature/2025.08.14

Ambala-gowthami:


pom.xml
```diff
@@ -17,7 +17,7 @@
 
   <properties>
 
-    <!-- Generic properties -->
+    <!-- Generic properties -1 -->
```
andrzejsydor, line 20: why such a change?

andrzejsydor (changes requested):
imo, useless changes

andrzejsydor (changes requested):
please close the PR

PR #2016, opened 2025-09-02
Upgrade to Spring Boot 4.0.0-M3 and add JSpecify annotations

mhalbritter:
This PR updates Spring Petclinic to use Spring Boot 4.0.0-M2 and it also adds JSpecify nullability annotations. NullAway is integrated in the build to get null checking at compile time. This has been done for the Maven and for the Gradle build.

NullAway has to be used with JDK 24 because it relies on a bugfix for type-use annotations which hasn't been backported to JDK 17 yet. The build must be run with JDK 24, however, the `-release` parameter for javac has been set to 17, so the resulting jar file is runnable with Java 17.

The gradle build doesn't work at the moment because of a Spring Data AOT repositories failure, I've opened https://github.com/spring-projects/spring-data-commons/issues/3354 for that.


.mvn/jvm.config
mhalbritter, line 1: This file is needed for errorprone, see https://errorprone.info/docs/installation.

PR #2027, opened 2025-09-09
assignment

ankursingh773:


src/main/java/org/springframework/samples/petclinic/model/PetTypeModel.java
```diff
@@ -0,0 +1,74 @@
+package org.springframework.samples.petclinic.model;
+
+
+public class PetTypeModel {
+
+	private String name,breed,skinType,skinColor,temperament;
+	private int height,lifespan;
+	private double weight;
+
+	public String getName() {
```
andrzejsydor, line 10: Just Lombok @Data can be here, without the setters and getters

src/main/java/org/springframework/samples/petclinic/owner/PetTypeController.java
```diff
@@ -0,0 +1,97 @@
...
+			PetType petTypeModel = petTypes.save(newPetType);
+			System.out.println(petTypeModel);
+
+			return ResponseEntity.badRequest().body("Pet Added");
+
+		} catch (Exception e) {
+			e.printStackTrace();
+		}
+
+		return null;
```
andrzejsydor, line 37: why null?

andrzejsydor: better use some 404 not found with message

PR #2029, opened 2025-09-09
Try-8

k-fathi:


Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+// test1
+// test2
+// test3
+// test4
+// test5
+// test6
+// test7
+// test8
+// test9
+// test10
```
andrzejsydor, line 66: can be removed

andrzejsydor (changes requested):
you don't need to create PR for tests purpose, even Draft is better

PR #2047, opened 2025-09-13
Feature v1

abdulwasay01-cloud:


.github/workflows/mavn.yaml
```diff
@@ -0,0 +1,17 @@
+---
+name: Manual maven steps for ci
+on:
```
andrzejsydor, line 3: what is the `mavn` shortcut about?

PR #2048, opened 2025-09-13
My first contribution

rehan525:
Added Lombok and dynamic server port option.

This is my first Open source contribution too.


src/main/resources/application.properties
```diff
@@ -21,5 +21,8 @@ logging.level.org.springframework=INFO
 # logging.level.org.springframework.web=DEBUG
 # logging.level.org.springframework.context.annotation=TRACE
 
+#Dynamic port
+server.port=0
```
andrzejsydor, line 25: why do you prefer dynamic port?

src/main/java/org/springframework/samples/petclinic/model/Person.java
```diff
@@ -18,13 +18,19 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
 
 /**
  * Simple JavaBean domain object representing an person.
  *
  * @author Ken Krebs
  */
 @MappedSuperclass
+@Data
```
andrzejsydor, line 31: are you sure that @Data is good option for entity ?

PR #2105, opened 2025-10-22
Change Java version from 25 to 21 in pom.xml

Mansisardana:
Updated the Java version in pom.xml from 25 to 21 
to match the system-installed JDK version (21).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
   <properties>
 
     <!-- Generic properties -->
-    <java.version>25</java.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -18,7 +18,7 @@
   <properties>
 
     <!-- Generic properties -->
-    <java.version>25</java.version>
+    <java.version>21</java.version>
```
andrzejsydor: Why downgrade?

PR #2135, opened 2025-11-24
Refactor: Fix code smells detected by PMD analysis

AleGTorres:
Hello! I ran a PMD analysis on the project and spotted a few code smells that could be cleaned up.

I've fixed 4 specific patterns to improve code consistency and performance:

* **Control Statement Braces:** Added missing braces to `if/else` blocks (e.g., in `PetController.java`) to prevent scope errors.
* **Guard Log Statement:** Wrapped logger calls with `isInfoEnabled()` checks to avoid unnecessary string processing.
* **Double Brace Initialization:** Replaced double-brace init with standard collection creation to avoid anonymous classes overhead.
* **Use Utility Class:** Added private constructors to utility classes to prevent instantiation.

I hope these contributions are helpful to the project. Thank you!

### DIFF
--- a/src/test/java/org/springframework/samples/petclinic/PostgresIntegrationTests.java
+++ b/src/test/java/org/springframework/samples/petclinic/PostgresIntegrationTests.java
@@ -110,7 +110,10 @@ public class PostgresIntegrationTests {
 
 		public void printProperties() {
 			for (EnumerablePropertySource<?> source : findPropertiesPropertySources()) {
-				log.info("PropertySource: " + source.getName());
+				if (log.isInfoEnabled()) {
+					log.info("PropertySource: " + source.getName());
+				}
+
 				String[] names = source.getPropertyNames();
 				Arrays.sort(names);
 				for (String name : names) {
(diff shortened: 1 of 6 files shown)


src/test/java/org/springframework/samples/petclinic/PostgresIntegrationTests.java
```diff
@@ -125,11 +128,16 @@ public void printProperties() {
...
+
 					if (resolved.equals(value)) {
-						log.info(name + "=" + resolved);
-					}
-					else {
-						log.info(name + "=" + value + " OVERRIDDEN to " + resolved);
+						if (log.isInfoEnabled()) {
+							log.info(name + "=" + resolved);
+						}
+						else {
```
dsyer: I think the braces fall in the wrong place here. Surely we can just put the `isInfoEnabled()` check around the whole of the code that used to be here?

src/test/java/org/springframework/samples/petclinic/PostgresIntegrationTests.java
```diff
@@ -125,11 +128,14 @@ public void printProperties() {
 					assertNotNull(sourceProperty.toString(), "source property toString() returned null.");
 
 					String value = sourceProperty.toString();
-					if (resolved.equals(value)) {
-						log.info(name + "=" + resolved);
-					}
-					else {
-						log.info(name + "=" + value + " OVERRIDDEN to " + resolved);
+
+					if (log.isInfoEnabled()) {
```
snicoll: Same, please revert.

snicoll (changes requested):
Some of the changes aren't useful IMO. The addition of missing braces is the best part of it.

PR #2144, opened 2025-12-05
Tests: Add null check test case to PetTypeFormatterTests

ronanpjr:
## Description
This PR improves the test coverage for the `PetTypeFormatter` class by adding a test case for handling `PetType` objects with null names.

## Changes
- Modified `PetTypeFormatterTests.java` to include `testPrintWithNullName`.
- Verifies that the formatter correctly returns `"<null>"` when the pet name is missing, rather than throwing a NullPointerException or returning an unexpected value.

## Context
While reviewing the `PetTypeFormatter`, I noticed that the behavior for printing a `PetType` with a null name was not explicitly tested. This ensures the formatter handles data anomalies gracefully.

## Checklist
- [x] Code compiles correctly
- [x] Created/updated tests
- [x] All tests passing locally
- [x] Overall Tests Branch and Line Coverage Improved

### DIFF
--- a/src/test/java/org/springframework/samples/petclinic/owner/PetTypeFormatterTests.java
+++ b/src/test/java/org/springframework/samples/petclinic/owner/PetTypeFormatterTests.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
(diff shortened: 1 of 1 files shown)


dsyer:
If it wasn't obviously just an exercise from a course I would merge this. Thank you for your effort, and good luck with your studies.

PR #2149, opened 2025-12-11
Fix: Add validation to prevent creating a pet with an empty name

Sagar-Sharma19-cmd:
This PR adds a server-side validation check in PetController to prevent creating a Pet with an empty name.

- Prevents invalid pet entries
- Provides clear error message to users
- Ensures consistent validation with the update form

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetController.java
@@ -105,6 +105,11 @@ class PetController {
 	@PostMapping("/pets/new")
 	public String processCreationForm(Owner owner, @Valid Pet pet, BindingResult result,
 			RedirectAttributes redirectAttributes) {
+		// Validate empty name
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/samples/petclinic/owner/PetController.java
```diff
@@ -105,6 +105,11 @@ public String initCreationForm(Owner owner, ModelMap model) {
 	@PostMapping("/pets/new")
 	public String processCreationForm(Owner owner, @Valid Pet pet, BindingResult result,
 			RedirectAttributes redirectAttributes) {
+		// Validate empty name
+		if (!StringUtils.hasText(pet.getName())) {
+			result.rejectValue("name", "required", "Pet name must not be empty");
```
dsyer: It looks like "required" already has a message associated with it, and it doesn't start with a capital letter - the third argument here is probably redundant?

PR #2177, opened 2025-12-25
Add validation to Owner telephone field

kriti20041:
This PR adds validation to the Owner.telephone field to ensure it is not blank
and contains a valid 10-digit number. This improves data integrity and user input validation.


andrzejsydor (changes requested):
the PR is not complete, no tests, no value, imo

PR #2180, opened 2025-12-26
Document owner telephone validation

afia0872:
Added a short documentation section explaining the owner telephone validation rule and where it is defined, to help new contributors understand existing validations.


snicoll:
Thanks but if we started doing this, there are many many many other things we'd have to document. This sort of stuff is best suited in the respective reference guide.

Issue #2157, opened 2025-12-15
Entity relationships should be modeled with fetch type LAZY


marschall:
Currently all entity relationships, especially `Owner`  are modeled as `EAGER`. This causes referenced `Pet`s, their `PetType`s and `Visit` to be loaded when a `Owner`  is loaded. This is suboptimal when searching for owners (http://127.0.0.1:8080/owners/find) which causes the following queries to be executed.

The actual search
```sql
SELECT
    o1_0.id,
    o1_0.address,
    o1_0.city,
    o1_0.first_name,
    o1_0.last_name,
    o1_0.telephone
  FROM owners o1_0
 WHERE o1_0.last_name LIKE ? ESCAPE '\'
 FETCH FIRST ? ROWS ONLY
```

for every result

```sql
SELECT
    p1_0.owner_id,
    p1_0.id,
    p1_0.birth_date,
    p1_0.name,
    t1_0.id,
    t1_0.name
  FROM pets  p1_0
  LEFT JOIN types t1_0 ON t1_0.id = p1_0.type_id
 WHERE p1_0.owner_id = ?
 ORDER BY p1_0.name;

SELECT
    v1_0.pet_id,
    v1_0.id,
    v1_0.visit_date,
    v1_0.description
  FROM visits v1_0
 WHERE v1_0.pet_id = ?
 ORDER BY v1_0.visit_date;
``` 

finally
```sql
SELECT COUNT(o1_0.id)
  FROM owners o1_0
 WHERE o1_0.last_name LIKE ? ESCAPE '\'
```

I assume this is done because we to avoid an N+1 issue because we display the pet names of an owner in the search results. It is not successful.

- model relationships as `LAZY`, possibly combined with `@BatchSize` or removal of the pet names from the search results
- use DTO projection for search results and use `listagg`

youssefbouraoui1:
@marschall can i tackle this

mohit-200:
Hi, I’d like to work on this issue. Please let me know if that’s okay.

PR #2197, opened 2026-01-08
Use LAZY fetch type for Owner pets relationship

Akashkashyap24:
Updated Owner entity to use FetchType.LAZY instead of EAGER for pets relationship to avoid unnecessary eager loading.
Fixes spring-projects/spring-petclinic#2157


dsyer:
No DCO. Assume superseded by #2204

PR #2202, opened 2026-01-10
Add Slovak localization messages_sk.properties

patodemjan:
Adds Slovak localization for PetClinic. German keys are preserved, only the values are translated into Slovak.


andrzejsydor (approved):

dsyer:
I can't see what went wrong in the failed build. Please re-submit if you still want to make this change.

PR #2211, opened 2026-01-17
Update pom.xml

cybergeekscloud:


pom.xml
```diff
@@ -46,6 +46,11 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-actuator</artifactId>
     </dependency>
+    <dependency>
+       <groupId>org.apache.logging.log4j</groupId>
+       <artifactId>log4j-core</artifactId>
```
andrzejsydor: Why do you need that?

cybergeekscloud: sorry, that was supposed to go to my branch, it was a test

PR #2228, opened 2026-01-31
Add /healthz and validate owner last name

MaximIgitov:
## Summary
- add /healthz endpoint with MockMvc coverage
- enforce non-blank owner lastName with clear validation message
- add tests for create/update validation cases

## Test plan
- not run (Java tooling not installed on server)

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/model/Person.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/Person.java
@@ -32,7 +32,7 @@ public class Person extends BaseEntity {
 	private String firstName;
 
 	@Column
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/samples/petclinic/model/Person.java
```diff
@@ -32,7 +32,7 @@ public class Person extends BaseEntity {
 	private String firstName;
 
 	@Column
-	@NotBlank
+	@NotBlank(message = "Last name is required")
```
andrzejsydor: Why do you need that?
What about firstName, etc?

PR #2280, opened 2026-03-08
Create Jenkinsfile

ashishj7744:


ashishj7744:
a

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM eclipse-temurin:17-jdk
```
andrzejsydor: why is needed if exists already?

PR #2310, opened 2026-03-26
Improved welcome page UI and added SCSS styling

DHARANISH9977:
Improved welcome page UI
 Added SCSS styling
 Enhanced layout and readability


DHARANISH9977 (commented):
Looks good!

dsyer:
I think this is in the category of "new feature" (so regretfully declined). Also we would need the compiled css I guess.

PR #2318, opened 2026-03-30
Add API to filter pets by type and name with lightweight response end…

shrutisharma-sh:
## Added Pet Filter API

- Added GET /api/pets to filter pets by type and name
- Added GET /api/pets/names for lightweight response (name + type only)
- Supports optional query parameters: type, name

This improves API usability for frontend and integrations.

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetRestController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetRestController.java
@@ -1,0 +1,75 @@
+package org.springframework.samples.petclinic.owner;
+
+import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/samples/petclinic/owner/PetRestController.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@RestController
+@RequestMapping("/api")
+
+public class PetRestController {
+
+	private final OwnerRepository ownerRepository;
```
ShashvatYadav, line 16: Use a service layer so that logic will be in the service layer rather than in the controller itself.

dsyer:
Seems like a new feature. We generally don't accept new features.

PR #2323, opened 2026-04-02
Added /livez and /readyz endpoints using Spring Boot Actuator health probes

GaneshReddy239:
- Delegate health checks to Actuator
- Expose only status (UP/DOWN)
- Enable health probes for liveness and readiness support


dsyer:
Duplicate (these endpoints are enabled by default in k8s I think).

PR #2350, opened 2026-04-13
Fix/add size validation person names

ThomasSchmidt1982:
Prevent database errors when users enter names exceeding the 30-character column limit. The @Size annotation triggers Bean Validation before persisting, displaying a proper form error instead of an error page.

Problem: Entering a first or last name longer than 30 characters in the Owner form causes a database error page instead of a user-friendly validation message.
How to reproduce: Go to /owners/new, enter more than 30 characters in the Last Name field, submit.
Fix: Added @Size(max = 30) constraint on firstName and lastName in Person.java to align with the VARCHAR(30) column definition in schema.sql.
Also added (length=30) to @Column to align JPA Mapping with VARCHAR(30) definition in scheme.sql


dsyer:
Please squash down to one commit and force push to re-submit.

PR #2368, opened 2026-04-20
Fix: visit date validation fix

nithyashree88:
- Added frontend validation using min attribute in date input
- Added backend validation in VisitController
- Prevents selecting past and current dates

Tested:
- Past date → shows error
- Future date → works correctly


nithyashree88:
I have fixed! please find my PR here #2368

Issue #2379, opened 2026-04-26
[BUG] Invalid 'page' parameter causes 500 error in VetController  and OwnerController


Abdu11ahMamun:
### Summary
Passing `page=0` or any negative value to paginated endpoints causes 
an unhandled `IllegalArgumentException` (500 Internal Server Error).
Passing a `page` value beyond total pages silently returns an empty 
result with no user feedback. Both `VetController` and 
`OwnerController` are affected.

---

### Affected Endpoints

| Endpoint | page=0 | page=999 |
|---|---|---|
| `/vets.html` | 500 error | Empty table, no feedback |
| `/owners` | 500 error | Not applicable (isEmpty check exists) |

---

### Steps to Reproduce

**Bug 1 — VetController (page=0):**
1. Run the application
2. Navigate to `http://localhost:8080/vets.html?page=0`
3. Result: 500 — "Page index must not be less than zero"

**Bug 2 — VetController (page beyond range):**
1. Navigate to `http://localhost:8080/vets.html?page=999`
2. Result: Page loads with empty vet table and no error message.
   Pagination shows [1 2] but URL remains `?page=999`.

**Bug 3 — OwnerController (page=0):**
1. Navigate to `http://localhost:8080/owners?page=0&lastName=`
2. Result: 500 — "Page index must not be less than zero"

---

### Root Cause

Both controllers use identical unguarded pagination logic:

```java
// VetController.findPaginated()
Pageable pageable = PageRequest.of(page - 1, pageSize);

// OwnerController.findPaginatedForOwnersLastName()
Pageable pageable = PageRequest.of(page - 1, pageSize);
```

When `page=0`: `PageRequest.of(-1, 5)` throws 
`IllegalArgumentException` — unhandled, results in 500.

When `page=999`: Spring Data returns an empty `Page` object 
silently — no redirect, no error message shown to user.

---

### Stack Trace

java.lang.IllegalArgumentException: Page index must not be less than zero
at AbstractPageRequest.<init>(AbstractPageRequest.java:49)
at PageRequest.of(PageRequest.java:77)
at VetController.findPaginated(VetController.java:65)
at VetController.showVetList(VetController.java:49)

Same trace reproduced at:
at OwnerController.findPaginatedForOwnersLastName(OwnerController.java:132)
at OwnerController.processFindForm(OwnerController.java:104)

---

### Expected Behavior
- `page <= 0` → redirect to first page
- `page > totalPages` → redirect to last valid page

[...]

PR #2417, opened 2026-05-14
[BUG FIX] Invalid 'page' parameter causes 500 error in VetController and OwnerController #2379

Suryansh2602:
Simple Explanation of What We Fixed
Here's what the problem was and how we fixed it in plain English:
The Problem 🐛
When users tried to access paginated pages (lists split into multiple pages) in the Petclinic app, two bad things happened:
Bad Page Numbers (0 or negative) → The app crashed with a 500 error
Example: Visiting /vets.html?page=0 or /vets.html?page=-5 broke the app
Too High Page Numbers → The app showed an empty table with no explanation
Example: Visiting /vets.html?page=999 (when there are only 3 pages) silently showed nothing
Why It Happened ⚠️
The code was doing basic math without checking if the numbers made sense:
page - 1 = array index
When page=0: 0-1 = -1 (negative index) → CRASH!
When page=999: Spring tried to fetch page 999, found nothing, and silently returned empty

The Fix ✅
We added boundary checks - like a bouncer checking if you're in the right place:
If page ≤ 0: Automatically redirect to page=1 (the first page)
If page is too high: Automatically redirect to the last valid page
Think of it like an elevator:
If someone enters floor -5 or 0, send them to floor 1 instead
If someone asks for floor 999 but the building only has 5 floors, send them to floor 5
What Gets Better 🎯
Users get helpful redirects instead of crashes or empty screens
Search filters are preserved when redirecting (e.g., if searching for "Franklin", the redirect keeps that search)
Better error handling - the app is now more robust
That's it! We're making the app smarter about handling page numbers that don't make sense. 📄

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/OwnerController.java
@@ -58,7 +58,7 @@ class OwnerController {
 
 	@InitBinder
 	public void setAllowedFields(WebDataBinder dataBinder) {
-		dataBinder.setDisallowedFields("id");
+		dataBinder.setDisallowedFields("id", "*.id");
 	}
 
 	@ModelAttribute("owner")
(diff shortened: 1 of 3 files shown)


dsyer:
Related issue #2379 was closed. This is a duplicate.

PR #2481, opened 2026-05-27
Register native resource hints for nested db/{h2,mysql,postgres}/ scripts

xp-vit:
The existing hints.resources().registerPattern("db/*") only matches files directly under db/, but every SQL script in src/main/resources/db/ lives one level deeper (db/h2/schema.sql, db/postgres/schema.sql, db/mysql/schema.sql, plus their data.sql peers). A GraalVM native build therefore boots, answers /actuator/health 200, and then 500s every business endpoint because the schema was never bundled into the image (the missing-pattern manifests as a "table not found" at first JDBC use).

Adding registerPattern("db/*/*") covers the nested layout so all three profiles (h2, mysql, postgres) work in native mode. The existing top-level "db/*" line is kept untouched for backward compatibility.

Reproduced on GraalVM Community 25 with Spring Boot 4.0.x and the postgres profile.


dsyer:
Build error is obvious. Please rebase and force push when you fix it.

PR #2482, opened 2026-05-28
display owners list on find page

kanastasov:


dsyer:
Seems like an unecessary change

PR #2336, opened 2026-04-07
Remove unused Vets object from VetController showVetList method

marcosgdz03:
## 🐛 Fix unused `Vets` object in VetController

Closes #2333

---

## 📌 Description

This PR removes an unused `Vets` object from the `showVetList` method in `VetController`.

Previously, a `Vets` instance was created and populated but never used or added to the model:

```java
Vets vets = new Vets();
Page<Vet> paginated = findPaginated(page);
vets.getVetList().addAll(paginated.toList());
return addPaginationModel(page, paginated, model);

PR #2483, opened 2026-05-28
Fix/unused vets object

kanastasov:
This removes an unused Vets object in VetController#showVetList.

The object was created and populated but never added to the model or returned,
making it dead code after pagination refactor.

The method already uses Page<Vet> and addPaginationModel for rendering,
so Vets is no longer needed.

No functional changes.


dsyer:
Duplicates #2336

PR #2491, opened 2026-06-01
Remove unused Vets object from VetController

king-407:
## Summary

- remove the unused `Vets` object from `VetController#showVetList`
- preserve existing license header, author tags, and JSON endpoint behavior

## Related issue

Closes #2333

## Testing

- `./mvnw -q -Dtest=VetControllerTests test`
- `./mvnw -q test`
- `git diff --check`

PR #2514, opened 2026-06-10
Remove unused Vets object from showVetList

Mani-dev1999:
## Summary

Removed the unused `Vets` object from `VetController#showVetList()`.

The object was created and populated but never added to the model or otherwise used. The page already relies on the paginated `listVets` model attribute provided by `addPaginationModel()`.

## Validation

* Application starts successfully
* `/vets.html` displays the vet list correctly
* Maven tests pass successfully


dsyer:
Duplicates #2491

PR #2531, opened 2026-06-16
Adds @NotNull to Visit.date plus a regression test. See description. 

SPM-Technologies:
Added @NotNull to Visit.date so visits cannot be saved without a date,
matching the existing validation style on Owner/Visit fields. Add a
VisitControllerTests case asserting the form is rejected when the date
is missing."


dsyer:
Too many AI artifacts. Seems like a borderline new feature too. Sorry.

PR #2535, opened 2026-06-17
Fix: add minimum length validation of 2 characters for firstName

Gopikatla:
Problem: First name field was accepting single character names 
like "A" which are not valid names.

Fix: Added min=2 to @Size annotation in Person.java
@Size(min = 2, max = 30, message = "First name must be between 2 and 30 characters")

Testing:
- "A" → shows error "First name must be between 2 and 30 characters" ✅
- "Hi" → saves successfully ✅
- "John" → saves successfully ✅


dsyer:
Seems like a new feature. Also DCO check fails.

PR #2557, opened 2026-06-30
Add validation to prevent numeric-only names for Owner and Pet

### LINKED ISSUE #2554: [Bug] Owner and Pet name fields accept numeric-only/special characters values, and long strings with no validation
# Bug: Missing validation for Owner and Pet names allows numeric-only values

## Labels
- `bug`
- `good-first-issue`

---

## Current Behavior

Both **Owner** and **Pet** name fields only have `@NotBlank` validation (checking for empty fields). There is no validation for:

- Names that are entirely numeric (e.g., `"123"`, `"12345"`)
- Names with only special characters (e.g., `"@#$%"`)
- Names with mixed invalid patterns

Users can create:

- Owners with names like `"123"`, `"987654"`, `"123 laghari"`
- Pets with names like `"456"`, `"9999"`, `"12345"`

This leads to **nonsensical data** in the system.

---

## Expected Behavior

Owner and Pet names should:

- Contain **at least one alphabetic character** (`A-Z` or `a-z`)
- Have a **reasonable maximum length** (e.g., **255 characters**)
- Show **clear validation error messages**
- Be **consistent** across both entities (**Owner** and **Pet**)

---

### For Owner

1. Go to **"Find Owners"** → click **"Add Owner"**
2. Enter `123` as the **First Name** (or any numeric-only name)
3. Fill other fields (**address, city, telephone**)
4. Click **"Add Owner"**
5. Observe the owner is saved with the numeric name

### For Pet

1. Go to an owner's page → click **"Add New Pet"**
2. Enter `456` as the pet name
3. Fill other fields
4. Click **"Add Pet"**
5. Observe the pet is saved with the numeric name

---

## Environment

- **Spring Boot version:** `4.1.0`
- **Java version:** `21` (or `17`)
- **Browser:** Any

---

## Affected Files

- `src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java` (lines **33–35**)
- `src/main/java/org/springframework/samples/petclinic/owner/Owner.java` *(extends `NamedEntity`)*
- `src/main/java/org/springframework/samples/petclinic/pet/Pet.java` *(extends `NamedEntity`)*

---

## Possible Fix (for future reference)

Add validation at the **NamedEntity** level (since both extend it):

```java
@NotBlank(message = "Name is required")
@Size(max = 255, message = "Name cannot exceed 255 characters")
@Pattern(regexp = ".*[a-zA-Z].*", message = "Name must contain at least one letter")
private String name;
```

---

## Screenshot

Image

Mohit040905:
## Summary
Fixes #2554 — Owner and Pet names could previously be entirely numeric (e.g. "123") or contain no letters at all, since the existing validation only checked for blank/empty values.

## Changes
- **Person.java**: Added `@Pattern(regexp = ".*[a-zA-Z].*")` to `firstName` and `lastName` fields, requiring at least one alphabetic character.
- **PetValidator.java**: Added a manual letter-check (`name.matches(".*[a-zA-Z].*")`) alongside the existing blank check, since Pet uses a custom `Validator` implementation rather than Bean Validation annotations.
- **NamedEntity.java**: Also added the same validation here for consistency, since other entities may extend this class in the future.

## Note on investigation
The issue description originally pointed to `NamedEntity.java` as the fix location. While investigating, I found that `Owner` actually extends `Person` (not `NamedEntity`) for its name fields, and `Pet` uses a separate `PetValidator` class with manual validation logic that bypasses Bean Validation annotations entirely. Both of these needed separate fixes to fully resolve the issue.

## Testing
Manually tested both flows:
- Adding an Owner with a numeric-only first/last name now shows a validation error
- Adding a Pet with a numeric-only name now shows a validation error
- Normal names with letters (including ones with numbers mixed in, e.g. "123 Smith") continue to work as expected

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/NamedEntity.java
@@ -18,6 +18,8 @@ package org.springframework.samples.petclinic.model;
 import jakarta.persistence.Column;
 import jakarta.persistence.MappedSuperclass;
 import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Pattern;
+import jakarta.validation.constraints.Size;
 
 /**
  * Simple JavaBean domain object adds a name property to <code>BaseEntity</code>. Used as
@@ -31,7 +33,9 @@ import jakarta.validation.constraints.NotBlank;
 public class NamedEntity extends BaseEntity {
 
 	@Column
-	@NotBlank
+	@NotBlank(message = "Name is required")
+	@Size(max = 255, message = "Name cannot exceed 255 characters")
+	@Pattern(regexp = ".*[a-zA-Z].*", message = "Name must contain at least one letter")
 	private String name;
 
 	public String getName() {
--- a/src/main/java/org/springframework/samples/petclinic/model/Person.java
+++ b/src/main/java/org/springframework/samples/petclinic/model/Person.java
@@ -19,6 +19,7 @@ import jakarta.persistence.Column;
 import jakarta.persistence.MappedSuperclass;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.Size;
+import jakarta.validation.constraints.Pattern;
 
 /**
  * Simple JavaBean domain object representing an person.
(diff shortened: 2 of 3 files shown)


dsyer:
DCO fails

PR #2579, opened 2026-07-18
i18n: add Japanese translation

CodesWithRobi:
Adds messages_ja.properties with all 52 keys matching existing translations.


dsyer:
Thank you

PR #2581, opened 2026-07-19
Add /hello endpoint and customize welcome text for local practice.

aslt918:


src/main/java/org/springframework/samples/petclinic/system/WelcomeController.java
```diff
@@ -27,4 +28,10 @@ public String welcome() {
 		return "welcome";
 	}
```
Mohrip, line 30: Either remove the /hello handler or make it non production (for example, guard it with a profile or conditional property) and add a short test or note in the PR explaining the intent

Mohrip (changes requested):
Either remove the /hello handler or make it non production (for example, guard it with a profile or conditional property) and add a short test or note in the PR explaining the intent

PR #2586, opened 2026-07-22
issue #2584: fix relation "unique_owner_pet_name" already exists in P…

walker99011:
#2584

PR #2595, opened 2026-07-28
Add IF NOT EXISTS to index creation statements in Postgres schema (fi…

### LINKED ISSUE #2584: [Bug] Relation "unique_owner_pet_name" already exists in PostgreSQL
## Description
The application fails to start after the initial run, throwing a PSQLException: ERROR: relation "unique_owner_pet_name" already exists when the PostgreSQL database has already been initialized.

The `application-postgres.properties` file contains the following setting:
```properties
# SQL is written to be idempotent so this is safe
spring.sql.init.mode=always
```
With the current migrations, this property causes two problems:
- New indexes are created on every application startup.
- The application fails to restart because a unique index that already exists cannot be created again.

## Steps to Reproduce
1. Start a PostgreSQL database using Docker Compose.
2. Activate the postgres profile.
3. Start spring-petclinic.
4. Shut down spring-petclinic.
5. Start spring-petclinic again.

## Expected Behavior
The application should start successfully without errors and without new indexes creation on each startup.

## Actual Behavior
The application fails to start with the error: `PSQLException: ERROR: relation "unique_owner_pet_name" already exists`.

## Proposed Solution
Add `IF NOT EXISTS` to all index creation statements.

Hemant1997:
Fixes #2584

Adds `IF NOT EXISTS` to all `CREATE INDEX` statements in the Postgres
schema.sql. Previously these ran unguarded on every startup (since
spring.sql.init.mode=always), causing a PSQLException on restart when
the unique index unique_owner_pet_name already existed.

Two previously unnamed indexes (on pets.name and pets.owner_id) were
given explicit names, since CREATE INDEX IF NOT EXISTS requires a
name to check existence against in PostgreSQL.

Verified locally: started Postgres via Docker Compose, ran the app
with the postgres profile active, stopped it, and restarted — no
errors on the second run.


dsyer:
I can't review it if the workflows fail.

dsyer:
Duplicates #2586

PR #2604, opened 2026-08-05
Show pet age on owner details page

mareklod33:
Adds Pet.getAge(), computed from birthDate as whole years via java.time.Period, and displays it next to Birth Date on the owner details page. Translation key added to the base messages file and all locale files (messages_en is intentionally left untranslated as the fallback). Includes unit tests for the age calculation, including the not-yet-had-birthday-this-year and null-birthDate edge cases.


dsyer:
Thanks. New features generally are not accepted though.

PR #2609, opened 2026-08-09
Fix owner addPet and pet validation

waghmaretejas:
Fixes the Owner.addPet() logic so persisted pets are added when they are not already associated with the owner.

Adds pet name length validation to reject names longer than 30 characters.

Added unit tests covering persisted pet association, duplicate prevention, and pet name length validation.

All existing tests pass.


dsyer:
Duplicates #2611

PR #2611, opened 2026-08-11
Fix Owner.addPet for persisted pets and validate pet names

### LINKED ISSUE #2600: Owner.addPet silently ignores persisted pets and PetValidator lacks name length check
The Owner.addPet(Pet pet) method in Owner.java contains a logic bug where it only adds a pet to the owner's pet list if pet.isNew() returns true (i.e., the pet has no ID). This means that any persisted pet (one that already has a database ID assigned) is silently ignored when added to an owner.

This bug manifests in PetController.updatePetDetails() where the else branch calls owner.addPet(pet) for a pet with an existing ID. Since the pet is not new, the call silently does nothing, and the pet is never associated with the owner.

Expected behavior: The addPet method should add any pet that is not already present in the owner's pet list, regardless of whether the pet is new or persisted.

Actual behavior: Only pets with a null ID are added. Pets with a valid ID are silently dropped.

Additionally, the PetValidator class lacks a name length validation check. The Pet entity inherits from NamedEntity which has a @Column(length=30) constraint, but the validator does not enforce this at the form validation level. This means names exceeding 30 characters pass validation but fail at the database level with a hard-to-debug constraint violation.

Steps to reproduce:

Create an Owner instance
Create a Pet instance and assign it an ID (e.g., pet.setId(5))
Call owner.addPet(pet)
Observe that owner.getPets() returns an empty list
Acceptance criteria:

Owner.addPet adds persisted pets that are not already in the list
Owner.addPet does not add duplicate pets
PetValidator rejects pet names longer than 30 characters
All existing tests continue to pass
New unit tests cover the fixed behavior
Word count: ~240 RULE 7 status: PASS

Zaid1681:
**Fixes #2600** by updating Owner.addPet to correctly associate persisted pets with an owner and by enforcing the pet name length constraint during validation.

**What changed**

- Updated Owner.addPet to add non-null pets when they are not already associated with the owner, including pets that already have an ID.
- Prevented duplicate pets from being added to the same owner.
- Extended PetValidator to reject pet names longer than 30 characters, matching the database column size.

**Added regression tests covering:**

- persisted pet association,
- duplicate pet prevention,
- long-name validation.

**Testing**
Verified with:
`mvn -q -Dtest=OwnerTests,PetValidatorTests test`

### DIFF
--- a/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/Owner.java
@@ -95,9 +95,18 @@ public class Owner extends Person {
 	}
 
 	public void addPet(Pet pet) {
-		if (pet.isNew()) {
-			getPets().add(pet);
+		if (pet == null) {
+			return;
 		}
+		if (getPets().contains(pet)) {
+			return;
+		}
+		for (Pet existingPet : getPets()) {
+			if (!existingPet.isNew() && !pet.isNew() && Objects.equals(existingPet.getId(), pet.getId())) {
+				return;
+			}
+		}
+		getPets().add(pet);
 	}
 
 	/**
--- a/src/main/java/org/springframework/samples/petclinic/owner/PetValidator.java
+++ b/src/main/java/org/springframework/samples/petclinic/owner/PetValidator.java
@@ -33,6 +33,8 @@ public class PetValidator implements Validator {
 
 	private static final String REQUIRED = "required";
 
+	private static final int MAX_NAME_LENGTH = 30;
+
 	@Override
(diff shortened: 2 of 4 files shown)


Zaid1681:
Hi maintainers, just following up on this PR. It has been open for a couple of weeks, and I wanted to check if someone could take a look when you have a chance.

The PR addresses #2600 by fixing persisted pet association in `Owner.addPet` and adding the missing pet-name length validation, along with regression tests. The current checks are passing and there are no merge conflicts.

Thanks for your time!

andrzejsydor (approved):

Zaid1681:
Hi @dsyer , can you please review the PR 
Thankyou !

PR #2645, opened 2026-08-31
Handle missing resources and invalid pagination

zldkr:
- Missing owners/pets now return HTTP 404 instead of 500.
- Invalid pagination parameters such as page=0 or negative values now return HTTP 400.
- Added MVC regression tests for Owner, Pet, Visit, and Vet controllers.
- Verified with Checkstyle and the full test suite.


dsyer:
Seems reasonable. You need to sign the commit though.

